PR #17, opened 2017-09-11
EVG-1947 stop using italics in logkeeper

john-m-liu:


public/static/logkeeper.css
```diff
@@ -37,8 +37,7 @@ td.log pre {
   border-right:1px solid #ccc;
   cursor:pointer;
 }
-.global{
-  font-style:italic;
+.global{ // placeholder for global styles
```
bsamek: Use `/* */` for CSS comments. https://www.w3.org/TR/css-syntax-3/#comment-diagram

john-m-liu: Done

PR #18, opened 2017-09-18
EVG-1940 add lobster viewer to logkeeper

john-m-liu:


tychoish:
I think the icon should go in static not template, but it's good other than that

PR #22, opened 2017-11-21
EVG-2355: logging improvements

tychoish:
this is largely lifted directly from evergreen itself.


log.go
```diff
@@ -57,3 +73,92 @@ func (l *Logger) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
...
+	// finally to the file as specified.
+
+	if fn == "" {
+		sender = getSystemLogger()
+		if err = sender.SetLevel(send.LevelInfo{Default: level.Info, Threshold: level.Debug}); err != nil {
+			return nil, errors.Wrap(err, "problem setting level for local system sender")
+		}
+
+		senders = append(senders, sender)
+	} else if fn == "NONE" || fn == "SKIP" {
```
chasinglogic: I would remove this and change line 154 to `else if fn != "NONE" || fn != "SKIP"`. It would seem to be more idiomatic and removes this "blank" block.

tychoish: awesome. will do, thanks!

PR #23, opened 2017-12-11
EVG-2338 make lobster the default view

john-m-liu:


views.go
```diff
@@ -542,6 +537,9 @@ func (lk *logKeeper) viewTestByBuildIdTestId(w http.ResponseWriter, r *http.Requ
 		if emptyLog {
 			lk.render.WriteJSON(w, http.StatusOK, nil)
 		}
+	} else if len(r.FormValue("html")) == 0 {
```
richardsamuels: Are we ever using this form value anywhere? Maybe we should drop the old UI? Otherwise lgtm.

john-m-liu:
The only way is the "html" query string. From what I can tell there really isn't a reason to use the old view over lobster so I haven't added a direct way to get to it

tychoish:
can we have a "?legacy=true" option? I can easily see a case where lobster is slower or breaks for some logs?

john-m-liu:
That's the same as ?html=true. Or are you saying that we should call the parameter "legacy" rather than "html"?

PR #27, opened 2018-01-26
EVG-2547 create script to update lobster

john-m-liu:


tychoish:
also changing the working directory is a thing I tend to like to avoid, because it makes debugging harder.

john-m-liu:
Made those changes. Hopefully it's less of a nightmare now

buildscripts/update-lobster.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+# this script pulls the evergreen-ci/lobster repo and copies the files so that
+# they can be served by logkeeper
+# it should only be run from the base logkeeper directory
+set -o errexit
```
bsamek: it might be helpful to also `set -o xtrace`

buildscripts/update-lobster.sh
```diff
@@ -0,0 +1,44 @@
...
+LOGKEEPER_STATIC_DIR=public/static
+LOGKEEPER_JS_DIR=js
+LOGKEEPER_CSS_DIR=css
+LOGKEEPER_TEMPLATES_DIR=templates/lobster/build
+
+# clone lobster repo and change to it
+pushd $SCRIPTS_DIR
+rm -rf $LOBSTER_DIR
+git clone $LOBSTER_REPO $LOBSTER_DIR
+pushd .lobster-temp
```
bsamek: use `cd` here, so you only need to pop once

buildscripts/update-lobster.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# build lobster
+npm install
+npm run build
+
+# replace existing js/css files in logkeeper with the updated ones
+popd && popd
```
bsamek: only pop once after `cd` change on 23

buildscripts/update-lobster.sh
tychoish: also this is LOBSTER_DIR

tychoish: I'd actually leave the pushd, and rearrange the pushes so they're at the beginning of the blocks, and pops so their at the end.

buildscripts/update-lobster.sh
john-m-liu: done

buildscripts/update-lobster.sh
john-m-liu: went with Sam's suggestion

buildscripts/update-lobster.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# clone lobster repo
+pushd $SCRIPTS_DIR
+rm -rf $LOBSTER_DIR
+git clone $LOBSTER_REPO $LOBSTER_DIR
+# build lobster
+pushd $LOBSTER_DIR
+npm install
+npm run build
+popd && popd
```
tychoish: I'd just put them on seperate lines

PR #32, opened 2018-02-23
EVG-2820: use contemporary context system

tychoish:


requests.go
```diff
@@ -75,18 +74,18 @@ func (lk *logKeeper) checkContentLength(r *http.Request) *apiError {
 
 type ctxKey int
 
-const (
-	requestCtxKey ctxKey = iota
-)
+const requestCtxKey ctxKey = iota
```
bsamek: Does it actually make sense to set a single constant to iota?

tychoish: not at all, but it's reasonable.

PR #33, opened 2018-02-26
EVG-2820 clean up vendor directory, fix panic when posting logs

john-m-liu:


tychoish:
we don't need to vendor gorilla/context

tychoish:
and run make vendor-clean to avoid vendoring files that we don't need.

john-m-liu:
I made vendor-clean but left gorilla/context in

PR #34, opened 2018-03-14
EVG-2819: cleanup logkeeper logging

tychoish:


log.go
```diff
@@ -70,8 +70,8 @@ func (l *Logger) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
 
 	res := rw.(negroni.ResponseWriter)
 
-	if res.Status() >= 400 {
-		grip.Warning(message.Fields{
+	if res.Status() != 200 {
+		grip.Info(message.Fields{
```
john-m-liu: Should this be at a higher level? Or maybe differentiate 2xx from 4xx/5xx

log.go
```diff
@@ -70,19 +78,19 @@ func (l *Logger) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
...
+
+	grip.Log(getLevel(code), message.Fields{
+		"method":   r.Method,
+		"request":  reqID,
+		"path":     r.URL.Path,
+		"duration": time.Since(start),
+		"action":   "completed",
+		"status":   code,
+		"remote":   remote,
+		"outcome":  http.StatusText(res.Status()),
```
richardsamuels: tiny nit, but evergreen logging just uses that status code. I'd also find it easier to query splunk with that. Maybe change this? Or include both?

Also, reuse the code variable here.

log.go
```diff
@@ -35,6 +35,14 @@ func NewLogger() *Logger {
 	return &Logger{ids}
 }
 
+func getLevel(l int) level.Priority {
+	if l == 200 {
```
richardsamuels: Use http.StatusOk instead?

PR #36, opened 2018-04-17
EVG-3152 don't gzip responses

john-m-liu:


john-m-liu:
I did something wrong when changing the go version to 1.8. I'll fix that separately

PR #44, opened 2018-11-28
EVG-5556 business logic for logkeeper cleanup

ablack12:


schema.go
```diff
@@ -106,3 +110,71 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
...
+
+func findAndDeleteTests(db *mgo.Database) ([]Test, error) {
+	now := time.Now()
+	query := bson.M{
+		"started": bson.M{"$lte": now.Add(-deletePassedTestCutoff)},
+		"failed":  false,
+	}
+	tests := []Test{}
+
+	err := db.C("tests").Find(query).Sort("started").Select(bson.M{"_id": 1}).All(&tests)
```
ablack12: Per a comment from john, will test a new version of this that limits the number of tests returned to some constant since logkeeper is absurdly huge.

schema.go
```diff
@@ -106,3 +113,69 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
...
+	}
+	var testIDs []bson.ObjectId
+	err := db.C("tests").Find(query).Sort("started").Limit(maxTests).Distinct("_id", &testIDs)
+	if err != nil {
+		return nil, errors.Wrap(err, "error finding tests")
+	}
+	filter := bson.M{
+		"_id": bson.M{"$in": testIDs},
+	}
+	_, err = db.C("tests").RemoveAll(filter)
```
john-m-liu: I think using your original query here might be faster, because they'll all be cached and also we don't need to do a big $in

schema.go
```diff
@@ -106,3 +113,69 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
...
+	if err != nil {
+		return nil, errors.Wrap(err, "error deleting tests")
+	}
+	return testIDs, nil
+}
+
+func deleteLogsByTests(db *mgo.Database, ids []bson.ObjectId) (*mgo.ChangeInfo, error) {
+	query := bson.M{
```
john-m-liu: You can't do a query like this because it'll collection scan. You'll either need to pass the build ID as well or find some way to do a join here. The indices on the collection are:

	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "buildlogs.logs"
	},
	{
		"v" : 1,
		"key" : {
			"build_id" : 1,
			"test_id" : 1,
			"seq" : 1
		},
		"name" : "build_id_1_test_id_1_seq_1",
		"ns" : "buildlogs.logs"
	},
	{
		"v" : 1,
		"key" : {
			"started" : 1
		},
		"name" : "started_1",
		"ns" : "buildlogs.logs",
		"expireAfterSeconds" : NumberLong(31536000)
	},
	{
		"v" : 1,
		"key" : {
			"build_id" : 1,
			"started" : 1
		},
		"name" : "build_id_1_started_1",
		"ns" : "buildlogs.logs",
		"background" : true
	}

schema.go
```diff
@@ -106,3 +113,69 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
...
+		"test_id": bson.M{"$in": ids},
+	}
+	return db.C("logs").RemoveAll(query)
+}
+
+func deleteBuildsWithoutTests(db *mgo.Database) error {
+	builds := []LogKeeperBuild{}
+	err := db.C("builds").Find(bson.M{}).Select(bson.M{"_id": 1}).All(&builds)
```
john-m-liu: You can't read all builds into memory. For this I think you need to do an aggregation to find builds that have no tests.

units/crons.go
```diff
@@ -21,8 +25,30 @@ func StartCrons(ctx context.Context, remote, local amboy.Queue) error {
...
+func PopulateCleanupOldLogDataJobs() amboy.QueueOperation {
+	return func(queue amboy.Queue) error {
+		catcher := grip.NewBasicCatcher()
+
+		db := db.GetDatabase()
+		tests, err := logkeeper.GetOldTests(db, time.Now())
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		for _, test := range *tests {
```
tychoish: shouldn't need to dereference this.

units/cleanup_old_log_data.go
```diff
@@ -0,0 +1,100 @@
...
+		Status string
+	}{}
+
+	if err = json.Unmarshal(payload, &taskInfo); err != nil {
+		j.AddError(err)
+		return
+	}
+
+	if taskInfo.Status != "success" {
+		return
```
tychoish: the problem with this is that it means we're going to be doing this request a bunch of times for all the old failed tests. maybe we should set failed=true, and then add that to the query so that we don't reexamine old failed tests?

units/crons.go
ablack12: It won't let me range over a pointer to a slice without dereferencing. We could not return a pointer? Or we dereference I think.

schema.go
```diff
@@ -106,3 +113,28 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
...
+func GetOldTests(db *mgo.Database, now time.Time) (*[]Test, error) {
+	query := bson.M{
+		"started": bson.M{"$lte": now.Add(-deletePassedTestCutoff)},
+	}
+	tests := []Test{}
+	err := db.C("tests").Find(query).Sort("-started").Limit(maxTests).All(&tests)
+	if err != nil {
+		return nil, errors.Wrap(err, "error finding tests")
+	}
+	return &tests, err
```
tychoish: don't return pointers to slices, slices are basically already pointers so this isn't needed, and you can return nil for a slice.

schema.go
```diff
@@ -106,3 +113,28 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
...
+}
+
+func CleanupOldLogsByTest(db *mgo.Database, id bson.ObjectId) error {
+	err := db.C("tests").RemoveId(id)
+	if err != nil {
+		return errors.Wrap(err, "error deleting test")
+	}
+
+	_, err = db.C("logs").RemoveAll(bson.M{"test_id": id})
```
tychoish: should the collection names here be constants? key names? 

does it make sense to pass the ID as a string? what is the type in the struct? 

are we sure that this is logged/

units/crons.go
tychoish: not return a pointer.

units/crons.go
```diff
@@ -21,8 +25,30 @@ func StartCrons(ctx context.Context, remote, local amboy.Queue) error {
...
 }
+
+// Queue Population Tasks
+
+func PopulateCleanupOldLogDataJobs() amboy.QueueOperation {
+	return func(queue amboy.Queue) error {
+		catcher := grip.NewBasicCatcher()
+
+		db := db.GetDatabase()
+		tests, err := logkeeper.GetOldTests(db, time.Now())
```
tychoish: time now? or threshold time?

units/cleanup_old_log_data.go
tychoish: I don't see that change. 

It would also work to delete the build/tests accordingly.

units/cleanup_old_log_data.go
```diff
@@ -0,0 +1,104 @@
...
+	defer resp.Body.Close()
+
+	payload, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		j.AddError(err)
+		return
+	}
+
+	taskInfo := struct {
+		Status string
```
john-m-liu: You should add a `json:"status"` here to be explicit about the field name

schema.go
```diff
@@ -106,3 +113,34 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
 	}
 	return build, nil
 }
+
+func UpdateFailedTest(db *mgo.Database, id bson.ObjectId) error {
```
john-m-liu: Since we're considering the db to be global state, it's awkward to pass that around. I would drop the db parameter and look it up in each function that needs it

schema.go
```diff
@@ -106,3 +113,34 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
...
+	update := bson.M{"failed": true}
+	return db.C("tests").UpdateId(id, update)
+}
+
+func GetOldTests(db *mgo.Database, now time.Time) (*[]Test, error) {
```
john-m-liu: I don't see why now needs to be a parameter - you can always you time.Now right?

units/cleanup_old_log_data.go
ablack12: What do you mean? 
`logkeeper.UpdateFailedTest(db, j.testID)` is what sets failed=true, and failed=false has been added to the query.

schema.go
ablack12: I feel like I had a reason at the time but who knows now :)

units/cleanup_old_log_data.go
ablack12: I think it makes sense to let builds get cleaned up separately, since cleaning them up here would involve looping through every single build.

schema.go
ablack12: In the Log struct, test_id is: `	TestId  *bson.ObjectId `bson:"test_id"`
So i think it makes more sense not to pass this id as a string.
What do you mean are we sure that this is logged? As in is the testID in logs populated? That would be our only way to connect the test to the log.

units/crons.go
ablack12: Not the threshold time. But John is right this parameter isn't actually necessary anymore.

PR #47, opened 2018-12-19
EVG-5707: speed up delete rate by 5x

tychoish:


units/crons.go
```diff
@@ -33,7 +33,7 @@ func StartCrons(ctx context.Context, remote, local amboy.Queue) error {
 // Queue Population Tasks
 
 func PopulateCleanupOldLogDataJobs() amboy.QueueOperation {
-	const limit = 600
+	const limit = 1500
```
bsamek: This is not 5x.

PR #48, opened 2019-01-02
Improve Migration Queue

tychoish:


globals.go
```diff
@@ -0,0 +1,14 @@
+package logkeeper
+
+import "time"
+
+const (
+	CleanupBatchSize = 1500
+
+	AmboyInterval      = time.Minute
+	AmboyWorkersPerApp = 8
+	AmboyTargetNumJobs = CleanupBatchSize / 4
```
bsamek: should be 5 if you want to keep the number the same as before this PR

units/stats_amboy.go
```diff
@@ -0,0 +1,116 @@
...
+		r["stale"] = stale
+	}
+
+	recentErrors, err := reporter.RecentErrors(ctx, time.Minute, reporting.StatsOnly)
+	j.AddError(err)
+	if recentErrors != nil {
+		r["errors"] = recentErrors
+	}
+
+	grip.InfoWhen(len(r) > 2, r)
```
bsamek: not >= 2? (if pending, etc., always return 0 instead of nil, then this is fine)

PR #50, opened 2019-02-05
EVG-5780 cleanup log data by build rather than test

ablack12:


schema.go
```diff
@@ -29,7 +29,6 @@ type Test struct {
 	Started   time.Time              `bson:"started"`
 	Ended     *time.Time             `bson:"ended"`
 	Info      map[string]interface{} `bson:"info"`
-	Failed    bool                   `bson:"failed"`
```
tychoish: I think we should probably keep this here even if we don't use it for legacy reasons? maybe safe to make it omit empty though

schema.go
```diff
@@ -119,35 +119,35 @@ func findBuildByBuilder(db *mgo.Database, builder string, buildnum int) (*LogKee
...
 	}
-	return tests, err
+	return builds, err
 }
 
-func getOldTestQuery() bson.M {
+func getOldBuildQuery() bson.M {
 	return bson.M{
 		"started": bson.M{"$lte": time.Now().Add(-deletePassedTestCutoff)},
 		"failed":  false,
```
tychoish: I think we might need to make this a "doesn't exist or is false, as most builds won't have this field, and this is, therefore weird.)

ablack12: How do I go about making sure this is indexed?

ablack12: Won't failed default to false? This is the same assumption we made when we were searching by test.

bsamek: Here are the indexes
```
mongos> db.builds.getIndexes()
[
	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "buildlogs.builds"
	},
	{
		"v" : 1,
		"key" : {
			"started" : 1
		},
		"name" : "started_1",
		"ns" : "buildlogs.builds",
		"expireAfterSeconds" : NumberLong(31536000)
	},
	{
		"v" : 1,
		"key" : {
			"buildnum" : 1,
			"builder" : 1
		},
		"name" : "buildnum_1_builder_1",
		"ns" : "buildlogs.builds"
	}
]
mongos> db.logs.getIndexes()
[
	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "buildlogs.logs"
	},
	{
		"v" : 1,
		"key" : {
			"build_id" : 1,
			"test_id" : 1,
			"seq" : 1
		},
		"name" : "build_id_1_test_id_1_seq_1",
		"ns" : "buildlogs.logs"
	},
	{
		"v" : 1,
		"key" : {
			"started" : 1
		},
		"name" : "started_1",
		"ns" : "buildlogs.logs",
		"expireAfterSeconds" : NumberLong(31536000)
	},
	{
		"v" : 1,
		"key" : {
			"build_id" : 1,
			"started" : 1
		},
		"name" : "build_id_1_started_1",
		"ns" : "buildlogs.logs",
		"background" : true
	}
]
mongos> db.tests.getIndexes()
[
	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "buildlogs.tests"
	},
	{
		"v" : 1,
		"key" : {
			"started" : 1
		},
		"name" : "started_1",
		"ns" : "buildlogs.tests",
		"expireAfterSeconds" : NumberLong(31536000)
	},
	{
		"v" : 1,
		"key" : {
			"build_id" : 1,
			"started" : 1
		},
		"name" : "build_id_1_started_1",
		"ns" : "buildlogs.tests"
	}
]
```

tychoish:
this seems ok, but I can you just double check and make sure that the queries are appropriately supported by indexes.

Also might be good to make sure the tests pass.

ablack12:
The tests that aren't passing are lint and pre-existing--seems like they've been there for a while.

I'm not totally clear on how/what indexes to add. Do I have access to these databases?

PR #64, opened 2022-06-29
EVG-17145 usage stats

ybrill:
Log usage stats once every ten seconds from each app server. This will allow us to answer the following questions in real time:
- the number of builds and tests created per unit time
- statistics on the size of log appends
- the number of accesses to a build, all build logs, and test logs

This will be useful for 1) answering what the s3 storage will cost per unit time and 2) ongoing throughput monitoring ([EVG-17147](https://jira.mongodb.org/browse/EVG-17147)).
The ticket specifies we should also be able to answer "the number of tests per build". I don't have a good way to keep track of this with this model (the tests could be added on different app servers, for instance). We could get a historical answer from the db, though. That should be sufficient since we don't need to know this on an ongoing basis: the only reason we need to know this is to estimate the cost of log downloads per unit time. 

I tried to simplify what we did in cedar: there is only one processing goroutine that does all the reading and writing of the cache. Along the way, I renamed the db package to env to make it clearer (to someone coming from the evergreen ecosystem) that this is the global thing we hang our global state on. This made sense for this PR since it's not only for the db now.


views.go
```diff
@@ -524,6 +561,11 @@ func (lk *logKeeper) viewTestByBuildIdTestId(w http.ResponseWriter, r *http.Requ
 			lk.logErrorf(r, "Error rendering template: %v", err)
 		}
 	}
+
+	grip.ErrorWhen(
```
julianedwards, line 565: For the "accessing" stats should we log earlier, in case this errors out before the log can actually be accessed or there is a redirect? Or do we only care about logs that are successfully returned?

julianedwards (changes requested):
Nice job with the logging stats, it's really elegant! A few small things/questions.

views.go
```diff
@@ -284,14 +295,19 @@ func (lk *logKeeper) appendLog(w http.ResponseWriter, r *http.Request) {
...
 		if err != nil {
 			lk.logErrorf(r, "Error inserting logs entry: %v", err)
 			lk.render.WriteJSON(w, http.StatusInternalServerError, apiError{Err: err.Error()})
 			return
 		}
 	}
 
+	grip.ErrorWhen(
+		sometimes.Percent(statsErrorLogPercentage),
+		errors.Wrap(env.StatsCache().LogAppended(int(r.ContentLength)), "adding test log append to stats cache"),
```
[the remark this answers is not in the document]
ybrill: done

env/env.go
```diff
@@ -0,0 +1,112 @@
+package env
```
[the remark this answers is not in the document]
ybrill: done

env/stats_cache.go
```diff
@@ -0,0 +1,149 @@
...
+		"message":                     "usage stats",
+		"interval_ms":                 time.Since(s.lastReset).Milliseconds(),
+		"num_builds_created":          s.buildsCreated,
+		"num_tests_created":           s.testsCreated,
+		"num_appends":                 len(s.logMBs),
+		"num_builds_accessed":         s.buildsAccessed,
+		"num_all_build_logs_accessed": s.allBuildLogsAccessed,
+		"num_test_logs_accessed":      s.testLogsAccessed,
+	}
+	if len(s.logMBs) > 0 {
```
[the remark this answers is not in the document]
ybrill: thank you!
The truth is, a few of these methods panic if the slice is empty.

views.go
ybrill: I went back and forth on this. In the end I concluded that by putting them at the end we're spared double counting requests that are retried. Especially for the s3 cost-estimate it made sense to me to exclude retries since they often result from db connection weirdness and the like.

julianedwards: What about the redirects? Are we going to miss those?

ybrill: I think on the contrary, by waiting until the end we'll only count it once. If we were to log on redirects we would count it now and also when the browser loads lobster and lobster gets the log through logkeeper's API.

PR #70, opened 2022-07-13
EVG-17147 defer recovery

ybrill:

In #69 I neglected to recover a panic in the logger loop.
In the unlikely event it does panic we'll stop logging, but that seems better than killing the application entirely. [The alert](https://github.com/mongodb/grip/blob/ae59645b416e7b68c8789ef3b5d43a93f6b732f1/recovery/recovery.go#L187) should bring it to our attention so we can restart the application manually.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
+	"github.com/mongodb/grip/recovery"
 	"github.com/mongodb/grip/sometimes"
 	"github.com/pkg/errors"
 	"github.com/urfave/negroni"
@@ -122,6 +123,8 @@ func (l *Logger) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
 }
 
 func (l *Logger) responseLoggerLoop(ctx context.Context) {
+	defer recovery.LogStackTraceAndContinue("logger loop")
+
 	ticker := time.NewTicker(loggerStatsInterval)
 	defer ticker.Stop()
 


logger.go
```diff
@@ -122,6 +123,8 @@ func (l *Logger) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
 }
 
 func (l *Logger) responseLoggerLoop(ctx context.Context) {
+	defer recovery.LogStackTraceAndContinue("logger loop")
```
julianedwards, line 126: I think this should be more like:
```
	defer func() {
		if err := recovery.HandlePanicWithError(recover(), nil, "logger loop"); err != nil {
			grip.Error(message.WrapError(err, message.Fields{
				"message": "panic in logger loop",
			}))
		}
	}()
```

julianedwards (changes requested):

logger.go
ybrill: Is there an advantage to that particular function? It seemed to me that `LogStackTraceAndContinue` was more consistent with other cases of recovery around logkeeper. They will have mostly the same effect: they both log a message like
```go
grip.Alert(message.Fields{
	"message": "hit panic; adding error",
	"op": "logger loop",
	"panic": <panic_error>.Error(),
	"stack": <the stack...>,
})
```

It seems like the only difference is that `HandlePanicWithError` will also log a _second_ message.

julianedwards: You are right, idk why we don't use this everywhere it is simpler!

julianedwards (approved):

logger.go
ybrill: thanks! 😄

PR #72, opened 2022-07-26
EVG-17147 logger as router middleware

ybrill:

Because the middleware wasn't being added to the _router_ the request context didn't contain the route at this level. This PR moves the middleware into the router so it will be called along the [chain](https://github.com/gorilla/mux/blob/c889844abd3601217c96aabc4b2dd89f6d904c01/mux.go#L144) after the route is [added to the request context](https://github.com/gorilla/mux/blob/c889844abd3601217c96aabc4b2dd89f6d904c01/mux.go#L201).

I tested this on a spawnhost by starting the binary and curling an endpoint. The routes were logged correctly (and it didn't panic 😄).


logger.go
```diff
@@ -74,52 +74,54 @@ func NewLogger(ctx context.Context) *Logger {
 	return l
 }
 
-// ServeHTTP calls the next function and incorporates the response into its response cache.
+// Middleware returns a handler that incorporates the response into its response cache.
```
johndaniels, line 77: I don't know if this makes sense to change in this review, but we keep referring to the buffer channel as a "cache" and that caused me confusion because I think it's just a processing queue/buffer, right?

logger.go
```diff
@@ -74,52 +74,54 @@ func NewLogger(ctx context.Context) *Logger {
...
-		),
-	)
+		}()
+
+		next.ServeHTTP(rw, r)
+
+		grip.ErrorWhen(
+			sometimes.Percent(logErrorPercentage),
+			message.WrapError(
+				l.addToCache(rw, r),
```
johndaniels, line 117: This ends up in Splunk, right?

johndaniels: Specifically errors returned from addToCache do?

main/logkeeper.go
```diff
@@ -79,10 +78,10 @@ func main() {
 
 	catcher := grip.NewCatcher()
 	router := lk.NewRouter()
+	router.Use(logkeeper.NewLogger(ctx).Middleware)
 	n := negroni.New()
-	n.Use(logkeeper.NewLogger(ctx))              // includes recovery and logging
 	n.Use(negroni.NewStatic(http.Dir("public"))) // part of negroni Classic settings
-	n.UseHandler(gorillaCtx.ClearHandler(router))
+	n.UseHandler(router)
```
johndaniels, line 84: To confirm, we're removing the explicit ClearHandler because the Gorilla Mux that we're using as a router should be doing that already, right?

johndaniels (approved):

main/logkeeper.go
ybrill: Although it's not directly related, I removed the ClearHandler call because it was just a relic. It was added in [this commit](https://github.com/evergreen-ci/logkeeper/commit/a9a77e4160ff3e4e2284f9c655f08f70ddcaac38) to add the request id to the request context. In [this later commit](https://github.com/evergreen-ci/logkeeper/commit/d32925bf1576b601543d958129fe5b1c642607ac) we switched out the gorilla.Context in favor of the (then new) request.Context that was built into go, but we didn't remove this line that was clearing the gorilla context. So we're clearing just clearing [these empty maps](https://github.com/gorilla/context/blob/8559d4a6b87e4f517ec1846eb90a192b8748cc89/context.go#L15-L16) before every request...

logger.go
ybrill: Yes, 10% of the errors from addToCache will be logged to Splunk. So like if we're filling the buffer faster than we can process it we'll see [response buffer is full](https://github.com/evergreen-ci/logkeeper/blob/4c355d457549c3a3ff8b06eb84ed347b38f6ec99/logger.go#L183) messages [there](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Dlogkeeper%20metadata.level%20%3E%3D%2070%20message%3D%22adding%20response%20to%20buffer%22&sid=1658955422.3469948&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-30m%40m&latest=now).

logger.go
ybrill: My intention in this context is that it adds to the cache by way of the buffer. I omitted mention of the buffer, though, because I thought it wouldn't matter from the perspective of a consumer of the Middleware.

I changed the function name `addToCache` to `addToRequestBuffer` to reflect what it's really doing.

PR #73, opened 2022-07-27
EVG-17347: Draft commit of what sample data might look like.

johndaniels:
Draft of adding some sample data to logkeeper to use for S3 reading/writing testing. We use the _bucketdata directory as the "bucket" in local testing in Pail, and we have a folder in testdata that represents a hypothetical set of files in S3. This also adds a "storage" package with the pail code inside of it (not sure about the name there) where we can put all this code and these tests.


storage/storage_test.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+	})
+	return err
+}
+
+func makeTestStorage(t *testing.T, initDir string) Storage {
+	os.Mkdir("../_bucketdata", 0755)
+	copyDirectory(initDir, "../_bucketdata")
+	bucket, err := pail.NewLocalBucket(pail.LocalOptions{
+		Path:   "../_bucketdata",
```
ybrill, line 41: Do we need to copy the data into a new directory? Could the bucket just be pointed at `testdata`?

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/metadata.json
```diff
@@ -0,0 +1,14 @@
+{
```
ybrill, line 1: Would it work for the logs and metadata for each test to reside under a test id prefix? This could enable us to filter to just a singe test's logs without reading the metadata for each test.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
```diff
@@ -0,0 +1,11 @@
+2022-07-23T07:15:34.848Z First Test Log Line
```
ybrill, line 1: We discussed over Slack that perhaps the filename could use unix timestamps. 
This would allow us to parse them with `time.Unix()`

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
```diff
@@ -0,0 +1,11 @@
...
+2022-07-23T07:15:34.866Z [js_test:geo_max:CheckReplOplogs] New session started with sessionID: {  "id" : UUID("4983fd5c-898a-4435-8523-2aef47ce91f3") } and options: {  "causalConsistency" : false }
+2022-07-23T07:15:34.868Z [js_test:geo_max:CheckReplOplogs] Recreating replica set from config {
+2022-07-23T07:15:34.868Z [js_test:geo_max:CheckReplOplogs] \t"_id" : "rs",
+2022-07-23T07:15:34.868Z [js_test:geo_max:CheckReplOplogs] \t"version" : 5,
```
ybrill, line 5: maybe it would be more straightforward to save the log lines as a json 2d array? Resmoke sends them to us this way, so we could literally just proxy them to S3. 
I could also picture it would make it easier when we parse it on retrieval if the timestamp and log line are already broken down for us.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+    "_id": "5a75f537726934e4b62833ab6d5dca41",
+    "builder": "MCI_enterprise-rhel_job0",
+    "buildnum": 157865445,
+    "started": "2022-07-23T07:15:05.353Z",
+    "name": "MCI_enterprise-rhel #157865445",
+    "info": {
+      "task_id": "mongodb_mongo_master_enterprise_f98b3361fbab4e02683325cc0e6ebaa69d6af1df_22_07_22_11_24_37"
```
ybrill, line 8: does the task_id need to be within a nested object? Is the idea that we might want to add more fields in the future?

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/metadata.json
```diff
@@ -0,0 +1,14 @@
...
+    "_id": "62dba0159041307f697e6ccc",
+    "build_id": "5a75f537726934e4b62833ab6d5dca41",
+    "build_name": "MCI_enterprise-rhel_job0",
+    "name": "geo_max:CheckReplOplogs",
+    "command": "",
+    "started": "2022-07-23T07:15:33.967Z",
+    "ended": null,
```
ybrill, line 8: If we're only going to write this once when the test is created then we can't ever know when the test ended. I imagine we won't need to know the started time/ended time anyway if we won't need them for fetching/merging logs.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/metadata.json
```diff
@@ -0,0 +1,14 @@
...
+    "info": {
+      "task_id": "mongodb_mongo_master_enterprise_rhel_80_64_bit_multiversion_all_feature_flags_retryable_writes_downgrade_last_continuous_2_enterprise_f98b3361fbab4e02683325cc0e6ebaa69d6af1df_22_07_22_11_24_37"
+    },
+    "phase": "retryable_writes_downgrade_last_continuous_2-enterprise_0",
+    "seq": 1
```
ybrill, line 13: this field seems specific to the db based implementation where we write new chunks to the db starting from the test's sequence number.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
ybrill: I remember once we discussed the filename being a base64 encoded json object.
Is that on the table?

testdata/simple/5a75f537726934e4b62833ab6d5dca41/metadata.json
johndaniels: This was straight copied from the bson... I'll pull it out.

ybrill:
It makes sense to me to also add the test to .evg.yml and the makefile so it will run on the waterfall.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
johndaniels: I don't know if it makes a huge difference... It's marginally easier to stream an array of lines or a jsonl file than a single json array object. I do _like_ the idea of using a standard format here though.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
johndaniels: We could do it. What would we be base64 encoding?

storage/storage_test.go
johndaniels: I thought about this. My concern had been if we want to write to the bucket, it'd be easier just to designate a temporary folder for doing test operations. I also considered just using a literal temporary folder for this, if we think that would be better?

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/metadata.json
johndaniels: Oh yeah I think I screwed this up from the document specification, sorry.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
johndaniels: I guess I'd vote for either jsonl (one json array per line) or be fine with just doing a big json array and relying on a streaming JSON reader. Do you have preferences?

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/metadata.json
johndaniels: I can just ditch started/ended anyway since they aren't as useful for the new algorithm to get all of the "global" logs that occur during a test.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
johndaniels: How would you feel about a fixed width unix epoch in milliseconds?

storage/storage_test.go
johndaniels: Actually, I just realized that we _are_ going to need to copy the whole file structure, but we should use pail's synchronization logic to do it, since at some point we're going to need to upload the test files to S3 for testing purposes.

johndaniels: This is updated to upload via pail's "push" method.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/metadata.json
johndaniels: Should be in a new folder now.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
johndaniels: This uses the unix epoch in ms now.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/2022-07-23T07-15-34.848Z!2022-07-23T07-15-34.869Z
johndaniels: Updated to use jsonl. Added extensions to make life easier for future folks downloading and using these files and for editing locally.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/metadata.json
johndaniels: moved to the top level.

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/metadata.json
johndaniels: Removed started/ended

testdata/simple/5a75f537726934e4b62833ab6d5dca41/tests/metadata.json
johndaniels: Removed

johndaniels:
I added the tests to the evg.yml file and also fixed our makefile so that we can actually test subpackages.

johndaniels:
Updated scaffolding to match Cedar's log format.

storage/storage_test.go
ybrill: nice!

storage/storage_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"os"
+	"testing"
+
+	"github.com/evergreen-ci/pail"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func makeTestStorage(t *testing.T, initDir string) Storage {
+	os.RemoveAll("../_bucketdata")
```
ybrill, line 14: should we `require.NoError` on these too?

ybrill (commented):
I think we could also add the new package to [packages](https://github.com/evergreen-ci/logkeeper/blob/769d4c96a53c751ecf033bba49106b92248eaed7/makefile#L4) so it will get included in the [test](https://github.com/evergreen-ci/logkeeper/blob/769d4c96a53c751ecf033bba49106b92248eaed7/makefile#L59) target, for instance.

storage/storage_test.go
johndaniels: Yes.

ybrill (approved):

PR #76, opened 2022-07-28
EVG-17156: Basic infrastructure for fetching logs from S3.

johndaniels:
This uses log_iterators copied over from Cedar, and takes advantage
of logs having an identical data format. There are a few changes
from the Cedar code:

1. The log iterators now filter chunks based on having any intersection
with their time range as opposed to being contained within their
time range. This is the biggest functional change from cedar's
iterators.
2. the LogChunkInfo struct has been modified to have a Key metho
rather than a Key value, to make it simpler to preserve the build
and test ids for later filter.

Also to note:
1. It looks like pail doesn't allow listing only one "level" of S3
keys, which means we pretty much always end up listing every
child of a build key. I think this should be fine from a
performance standpoint anyway, since we shouldn't have more than
1000ish subkeys anyway.
2. This code explicitly avoids using the MergeIterator across
every log file, because we would like to be able to load log
files in parallel. Instead it uses the BatchIterator. In principle
a parallel merge should be possible, but it is unnecessary
to preserve the existing behavior, which currently does not
correctly handle overlapping test logs.


storage/storage.go
```diff
@@ -13,3 +20,120 @@ func NewStorage(bucket pail.Bucket) Storage {
...
+	sort.Slice(testChunks, func(i, j int) bool {
+		return testChunks[i].Start.Before(testChunks[j].Start)
+	})
+
+	testTimeRange := TimeRange{
+		StartAt: testChunks[0].Start,
+		EndAt:   testChunks[len(testChunks)-1].End,
+	}
+
+	testChunkIterator := NewBatchedLogIterator(storage.bucket, testChunks, 4, testTimeRange)
```
ybrill, line 115: why is it better to use the batched iterator as opposed to the parallelized iterator?

storage/storage.go
```diff
@@ -13,3 +20,120 @@ func NewStorage(bucket pail.Bucket) Storage {
...
+		}
+	}
+
+	sort.Slice(buildChunks, func(i, j int) bool {
+		return buildChunks[i].Start.Before(buildChunks[j].Start)
+	})
+	buildChunkIterator := NewBatchedLogIterator(storage.bucket, buildChunks, 4, testTimeRange)
+
+	// Merge everything together
+	return NewMergingIterator(testChunkIterator, buildChunkIterator), nil
```
ybrill, line 138: if we just have two streams do we need to use the heap implementation? (it would seem merging two streams has the advantage of being O(n+m) vs. the heap is O(n log n + m log m))

storage/iterator.go
```diff
@@ -0,0 +1,18 @@
+package storage
+
+import "context"
+
+// Iterator represents a cursor for generic iteration over a sequence of items.
+type Iterator interface {
```
ybrill, line 6: is it beneficial to have these methods in their own interface?

storage/storage.go
```diff
@@ -13,3 +20,120 @@ func NewStorage(bucket pail.Bucket) Storage {
...
+			testChunks = append(testChunks, chunks[i])
+		}
+	}
+	sort.Slice(testChunks, func(i, j int) bool {
+		return testChunks[i].Start.Before(testChunks[j].Start)
+	})
+
+	testTimeRange := TimeRange{
+		StartAt: testChunks[0].Start,
+		EndAt:   testChunks[len(testChunks)-1].End,
```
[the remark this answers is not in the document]
ybrill: I think I understand: this is also based on the simplifying assumption we made that there won't be overlapping logs, in this case test logs.

storage/storage.go
johndaniels: BatchLogIterator and the parallelized iterator are actually the same implementation, but the parallelized iterator downloads every chunk, while the batch iterator does the next "n" chunks based on the batch size once it has exhausted existing chunks. The ParallelizedLogIterator is just a BatchedLogIterator where n = len(chunks).

We'll presumably want to tune this for performance. If we're ok just always loading everything into memory than the ParallelizedLogIterator is probably faster. It's a little more complicated than that since neither one actually fully downloads the log chunks from the bucket. They both just open the HTTP connection and get the initial HTTP response using the .Get() method from pail, but I'm pretty sure that doesn't actually read in the whole body into memory.

storage/storage.go
johndaniels: We don't need to use the heap implementation, but it was already written to use the Iterator interface and has been tested in production code, which is nice.

As far as runtime complexity, that "log(N)" component is dependent on heap operations being logN. In the case of the MergingIterator, that is going to be proportional to the size of the heap, which is just the constant (2) since it is a heap of the two iterators supplied. That brings the algorithmic complexity down do O(n + m). Basically since we're not actually putting all the log chunks or log lines on a heap, the heap operation complexity should be constant, and not scale up, and we do one heap operation per log line, so constant O(c * (n + m) ) = O(n+m).

That said, it's a fair bit of additional fixed pointless computational overhead per log line, but I'm guessing that that is dwarfed by the overhead of network handling, processing strings, etc.

storage/iterator.go
johndaniels: It's nice to hide the implementation of Batched vs Merging iterator from the storage functions calling them. It's a pretty stable interface, based on it not having changed in Cedar in forever.

storage/log_chunk_info.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+// LogChunkInfo describes a chunk of log lines stored in pail-backed offline
+// storage.
+type LogChunkInfo struct {
+	BuildId  string
```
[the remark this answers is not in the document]
johndaniels: This is fixed. Also found a place where I inexplicably used snake_case.

storage/log_iterator.go
```diff
@@ -0,0 +1,655 @@
...
+	"context"
+	"fmt"
+	"io"
+	"runtime"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/evergreen-ci/logkeeper"
```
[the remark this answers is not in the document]
johndaniels: I pulled everything out into a "models" package, but I'm happy to drop that in favor of your work since all I did was copy stuff over to the "models" folder...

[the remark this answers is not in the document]
johndaniels: Also happy to get on a call and pair on this if it would be easier to hash out the design of the packages that way!

storage/time_range.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Check returns true if the given time is within the TimeRange (inclusive) and
+// false otherwise.
+func (t TimeRange) Check(ts time.Time) bool {
+	if (ts.After(t.StartAt) || ts.Equal(t.StartAt)) &&
+		(ts.Before(t.EndAt) || ts.Equal(t.EndAt)) {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
johndaniels: Removed.

storage/time_range.go
```diff
@@ -0,0 +1,52 @@
...
+		startAt = endTime.Add(-duration)
+	} else {
+		endTime = startAt.Add(duration)
+	}
+
+	return TimeRange{
+		StartAt: startAt,
+		EndAt:   endTime,
+	}
+}
```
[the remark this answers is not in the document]
johndaniels: Removed...

storage/storage.go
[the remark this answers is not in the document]
johndaniels: I added in code to handle that possibility anyway? But yeah that is I think the assumption the existing code makes.

storage/storage.go
ybrill: That sounds good to me.

I'd always assumed it was buffered somewhere on the client until we read from the reader. Is there a way for the client to throttle the server so it only sends when we're ready?

storage/storage.go
ybrill: oh wow! That makes sense.

.evg.yml
```diff
@@ -175,7 +175,7 @@ buildvariants:
   - name: ubuntu1804
     display_name: Ubuntu 18.04
     expansions:
-      build_env: "GOROOT=/opt/golang/go1.16 PATH=/opt/golang/go1.16/bin:$PATH"
+      build_env: "GOROOT=/opt/golang/go1.18 PATH=/opt/golang/go1.18/bin:$PATH"
```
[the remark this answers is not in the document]
ybrill: Why do we want to upgrade go version? Is it just to stay current?

storage/storage.go
johndaniels: I _think_ if you don't call read() what will happen is that the TCP buffer will fill up and the server will wait to send more bytes. I feel like I'd need to test this empirically though to be sure. It could be the case that ParalleizedLogIterator is fine in that case. It's an easy change to make any time though.

johndaniels: At a glance it looks like the TCP buffer is calculated using this formula: ```max(87380, min(4 MB, tcp_mem[1]*PAGE_SIZE/128))
```
In principle unless the go runtime is calling recv for us then that's around memory will be used until we read from the body.

Comment (johndaniels) [reply storage/storage.go:122]: I'm pretty sure in practice that means 4 MB.

File: .evg.yml (hunk 9 above)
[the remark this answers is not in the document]
Comment (johndaniels) [reply .evg.yml:178]: Mostly because I wanted to use UnixMilli... I can rewrite the code without it though.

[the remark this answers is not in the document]
Comment (johndaniels) [reply .evg.yml:178]: It _is_ nice to stay current and it wasn't too hard. I only ended up having to also update the linter and the few places where we use conditional build instructions.

File: storage/storage.go (hunk 1 above)
Comment (johndaniels) [reply storage/storage.go:122]: See https://github.com/johndaniels/gobuffertest

File: .evg.yml [hunk 10]
```diff
@@ -188,7 +188,7 @@ buildvariants:
   - name: race-detector
     display_name: Race Detector
     expansions:
-      build_env: "GOROOT=/opt/golang/go1.16 PATH=/opt/golang/go1.16/bin:$PATH"
+      build_env: "GOROOT=/opt/golang/go1.18 PATH=/opt/golang/go1.19/bin:$PATH"
```
ybrill, line 191: It looks like the path was intended to be 1.18 but is a typo

.evg.yml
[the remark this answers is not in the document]
ybrill: Copying over from Slack: I can't find where the code uses `UnixMilli`. If it does I'd think we definitely need to increment the go directive to get the build to work.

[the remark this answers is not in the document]
johndaniels: It was one of the unused functions that we deleted.

I'm going to revert the upgrade to 1.8 for now because we don't need it and it should be a separate PR.

ybrill (approved):

PR #77, opened 2022-07-28
EVG-17147 sort slice

ybrill:

The panics from the previous deploy were due to failing [this check](https://github.com/gonum/gonum/blob/1180659542815ddb51545f83860386272ebf0d9b/stat/stat.go#L504-L506). Looking over the entire list of checks there it looks like we're safe now.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -239,7 +239,7 @@ func sliceStats(sample, histogramBins []float64) message.Fields {
 	if len(sample) == 0 {
 		return message.Fields{}
 	}
-	sort.Float64Slice(sample).Sort()
+	sort.Float64s(sample)
 
 	min := sample[0]
 	max := sample[len(sample)-1]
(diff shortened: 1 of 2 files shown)


logger.go
```diff
@@ -238,9 +239,10 @@ func sliceStats(sample, histogramBins []float64) message.Fields {
 	if len(sample) == 0 {
 		return message.Fields{}
 	}
+	sort.Float64Slice(sample).Sort()
```
johndaniels, line 242: Wouldn't it be simpler to use https://pkg.go.dev/sort#Float64s?

johndaniels (changes requested):

johndaniels (approved):
I'm approving this because the requested change doesn't really matter.

logger.go
ybrill: totally!

ybrill: done

PR #78, opened 2022-07-29
EVG-17149 move db logic into models

ybrill:
There's still work to do to fix the tests and write new ones.
My question is: should we go with this larger overhaul or not?


johndaniels (changes requested):
Looks like a great plan! Branch doesn't currently compile for me but I support merging once it does!

templates/build.html
```diff
@@ -5,18 +5,18 @@
...
           {{if .Build.Info}}
-            {{if .Build.Info.task_id}}
-              (<a href="https://evergreen.mongodb.com/task/{{.Build.Info.task_id}}"> {{.Build.Info.task_id}}</a>) 
+            {{if .Build.Info.TaskID}}
+              (<a href="https://evergreen.mongodb.com/task/{{.Build.Info.TaskID}}"> {{.Build.Info.TaskID}}</a>) 
             {{end}}
           {{end}}
         </h2>
-      <h3><a href="https://evergreen.mongodb.com/lobster/build/{{StringifyId .Build.Id}}/all">Complete logs for all</a></h3>
+      <h3><a href="https://evergreen.mongodb.com/lobster/build/{{.Build.Id}}/all">Complete logs for all</a></h3>
```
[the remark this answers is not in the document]
ybrill: Currently the code is written to be agnostic to the type of the build ID. I believe the history there is that they wanted to use the build ID as a shard key for logs so they changed it to be a string. Since there was a mix of types in the db they made the code able to handle either type. Now, though, all the build documents with `ObjectId` ids (and other documents with references to build ids) have TTLed so we no longer need to handle it.

model/build.go
```diff
@@ -0,0 +1,138 @@
...
+	"gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+)
+
+const (
+	deletePassedTestCutoff = 30 * (24 * time.Hour)
+	buildsCollection       = "builds"
+)
+
+var oldBuildsQuery = bson.M{
```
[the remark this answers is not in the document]
ybrill: Thinking more about it, we only use it once now, so it made sense to me to just move the query into the function where it's used.

model/log.go
```diff
@@ -51,20 +51,20 @@ func RemoveLogsForBuild(buildID string) (int, error) {
...
 func findLogsInWindow(query bson.M, sort []string, minTime, maxTime *time.Time) chan *LogLineItem {
-	db, closeSession := db.DB()
-	defer closeSession()
-
 	outputLog := make(chan *LogLineItem)
 	logItem := &Log{}
 
 	go func() {
+		db, closeSession := db.DB()
+		defer closeSession()
```
johndaniels, line 69: Oof... good catch.

model/log.go
```diff
@@ -176,21 +176,21 @@ func GroupLines(lines []LogLine) ([]LogChunk, error) {
...
 			logChars = 0
 			chunks = append(chunks, currentChunk)
 			currentChunk = LogChunk{}
 		}
 
-		logChars += len(line.Msg)
-		currentChunk.Lines = append(currentChunk.Lines, line)
-		if currentChunk.EarliestTime != nil && line.Time.Before(*currentChunk.EarliestTime) {
-			currentChunk.EarliestTime = &line.Time
+		logChars += len(lines[i].Msg)
```
johndaniels, line 199: I'm not sure why we changed this to use the index instead?

model/log.go
```diff
@@ -176,21 +176,21 @@ func GroupLines(lines []LogLine) ([]LogChunk, error) {
...
+		currentChunk.Lines = append(currentChunk.Lines, lines[i])
+		if currentChunk.EarliestTime == nil || lines[i].Time.Before(*currentChunk.EarliestTime) {
+			currentChunk.EarliestTime = &lines[i].Time
```
johndaniels, line 202: I think this is fine because Time objects are immutable, but it did cause me some alarm to create a pointer to that Time value.

testutil/testutil.go
```diff
@@ -0,0 +1,45 @@
...
+// ClearCollections clears all documents from all the specified collections,
+// returning an error immediately if clearing any one of them fails.
+func ClearCollections(collections ...string) error {
+	db, closer := db.DB()
+	defer closer()
+
+	for _, collection := range collections {
+		_, err := db.C(collection).RemoveAll(bson.M{})
+		if err != nil {
+			return errors.Wrapf(err, "clearign collection '%s'", collection)
```
johndaniels, line 41: typo of "clearign".

testutil/testutil.go
```diff
@@ -0,0 +1,45 @@
...
+	return nil
+}
+
+// ClearCollections clears all documents from all the specified collections,
+// returning an error immediately if clearing any one of them fails.
+func ClearCollections(collections ...string) error {
+	db, closer := db.DB()
+	defer closer()
+
+	for _, collection := range collections {
```
johndaniels, line 38: If we wanted to be paranoid we could verify that the db we are talking to is "logkeeper_test" (to avoid accidentally deleting a real DB if an environment was misconfigured or something.

johndaniels (approved):

testutil/testutil.go
ybrill: fixed

testutil/testutil.go
ybrill: I tried a different strategy where we set a build tag during testing that will create a connection to the db and implement this function while production code doesn't implement the function. Apparently [gofmt](https://go.dev/doc/go1.17#gofmt) adds both types of build tags so it will be forwards and backwards compatible.

Does this sound like a good idea? One effect that occurs to me is that if a developer isn't using the makefile to run tests (like using VSCode's "Run Test" button) they'll need to set the `test` build flag (in workspace settings for vscode).

model/log.go
ybrill: I removed the whole earliest time thing since it seems like we already rely on the assumption that log lines are already sorted by time.

model/log.go
ybrill: This was related to what you pointed out in [the next thread](https://github.com/evergreen-ci/logkeeper/pull/78#discussion_r936974048): we were capturing a pointer to the `line.Time` on each iteration of the loop. Since the iterator variable is always the same the pointers were always to the same thing so all the chunks had the earliest time of the last chunk.
I think the whole idea was unwise and opaque. I've removed it.

testutil/testutil.go
johndaniels: I usually prefer to be explicit about setup but I think that matches how we do that in evergreen.

johndaniels (approved):

testutil/testutil.go
ybrill: I've backed out the changes to use a build tag for this.
I restored the explicit call to `InitDB()`

PR #79, opened 2022-08-03
EVG-17155: Add functions to get global build log lines and all log lines.

johndaniels:
This adds functions to get all the "global" log lines that don't have a test id, as well as all log lines, merged in appropriate order.


storage/storage.go
```diff
@@ -90,55 +90,131 @@ func (storage *Storage) getAllChunks(context context.Context, buildId string) ([
...
+		return nil, err
+	}
+
+	testChunks := testChunksWithId(allTestChunks, testId)
+
+	sortByStartTime(testChunks)
+
+	testTimeRange := TimeRange{
+		StartAt: testChunks[0].Start,
+		EndAt:   getLatestTime(testChunks),
```
ybrill, line 211: It seems like this could be different from the way the end time is currently determined. There the maxTime [is the start of the next test](https://github.com/evergreen-ci/logkeeper/blob/5957d4855c595454db0fc288df6ac2d2b6fb9e46/views.go#L625).
The difference would show if there's a pause between tests but there are ongoing global logs: today logkeeper returns global logs from the interim, whereas determining the end of the time range from the last test chunk will preclude them.

storage/storage.go
```diff
@@ -90,55 +90,131 @@ func (storage *Storage) getAllChunks(context context.Context, buildId string) ([
...
+	}
+
+	buildChunkIterator := NewBatchedLogIterator(storage.bucket, buildChunks, 4, timeRange)
+	testChunkIterator := NewBatchedLogIterator(storage.bucket, testChunks, 4, timeRange)
+
+	// Merge test and build logs
+	return NewMergingIterator(testChunkIterator, buildChunkIterator), nil
+}
+
+func (storage *Storage) GetGlobalLogLines(context context.Context, buildId string) (LogIterator, error) {
```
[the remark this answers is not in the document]
johndaniels: For some reason I thought this was important. We only get either "all" logs, or just the logs during a test then?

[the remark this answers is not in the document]
ybrill: I think so. [This route](https://github.com/evergreen-ci/logkeeper/blob/cb88682ab734bf7b06f86d67dae87c230df17b92/views.go#L512) gets all the logs for a build and [this one](https://github.com/evergreen-ci/logkeeper/blob/cb88682ab734bf7b06f86d67dae87c230df17b92/views.go#L513) gets the logs during a test.

storage/storage.go
johndaniels: This should be fixed. We technically get the beginning of whatever test chunk has a start time bigger than the end time of the latest chunk of the test. I think that essentially does the same thing assuming the tests are non-overlapping. If the tests are overlapping then the new logic would find some test chunk in the other test that was strictly after this test, which seems... fine. Although we are assuming that the tests are non-overlapping.

storage/storage.go
```diff
@@ -90,55 +90,132 @@ func (storage *Storage) getAllChunks(context context.Context, buildId string) ([
...
+
+	testTimeRange := TimeRange{
+		StartAt: testChunks[0].Start,
+		EndAt:   logEndTime,
+	}
+
+	testChunkIterator := NewBatchedLogIterator(storage.bucket, testChunks, 4, testTimeRange)
+
+	sortByStartTime(buildChunks)
+	// This batchedlogiterator will filter out buildChunks that don't intestect with testTimeRange
```
ybrill, line 218: "intestect" -> "intersect"

ybrill: Although they will be filtered out it seems like it could be more efficient if we only pass chunks that intersect with the time range, at least for later tests in the build, since the iterator gets each chunk before it filters it out.

storage/storage.go
```diff
@@ -90,55 +90,132 @@ func (storage *Storage) getAllChunks(context context.Context, buildId string) ([
...
+	buildChunks := []LogChunkInfo{}
+	for i := 0; i < len(chunks); i++ {
+		if chunks[i].TestID == "" {
+			buildChunks = append(buildChunks, chunks[i])
+		}
 	}
 
 	testChunks := []LogChunkInfo{}
 	for i := 0; i < len(chunks); i++ {
 		// Find our test id
```
ybrill, line 108: this is no longer the case

storage/storage.go
```diff
@@ -90,55 +90,132 @@ func (storage *Storage) getAllChunks(context context.Context, buildId string) ([
...
+	if err != nil {
+		return nil, err
+	}
+
+	sortByStartTime(buildChunks)
+	sortByStartTime(testChunks)
+
+	timeRange := TimeRange{
+		StartAt: minTime(buildChunks[0].Start, testChunks[0].Start),
+		EndAt:   maxTime(getLatestTime(buildChunks), getLatestTime(testChunks)),
```
ybrill, line 159: it would be nice if TimeRange had an option to not specify a min and/or max time. Like we don't really need a min or max time here: it's just to signal to the iterator not to filter anything out.

storage/storage.go
johndaniels: Sorry, what is?

ybrill: sorry: I meant that we're no longer finding _our_ test id, but rather any chunk with a test id

johndaniels: Oh the comment...

ybrill: yes. If we prefer I can omit pedantic things from my reviews.

storage/storage.go
johndaniels: They get pre-filtered [here](https://github.com/evergreen-ci/logkeeper/blob/main/storage/log_iterator.go#L199) by the iterator, before fetching anything.

johndaniels: I will fix the comment though.

ybrill: oh! I missed that

testdata/between/5a75f537726934e4b62833ab6d5dca41/1000000000301000000_1000000000302000000_2
```diff
@@ -0,0 +1,2 @@
+  0       100000000030Log301
```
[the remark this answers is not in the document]
johndaniels: Yes

storage/storage.go
johndaniels: I solved this by adding TimeRangeMin and TimeRangeMax. I looked into making it optional, but that makes a lot of the logic hard and I don't think we need to represent times outside of the range 1970 - 2200

testdata/between/5a75f537726934e4b62833ab6d5dca41/1000000000301000000_1000000000302000000_2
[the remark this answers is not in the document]
johndaniels: Wow, thanks for the thoroughness!

ybrill (approved):

storage/storage.go
johndaniels: Oh no, please don't

PR #80, opened 2022-08-05
EVG-17149 store build

ybrill:
[EVG-17149](https://jira.mongodb.org/browse/EVG-17149)

Add a method to `storage` to store build metadata in S3. As I added it it seemed to me that `storage.go` had reached critical mass and would benefit from decomposition, so I attempted to break it up. The merge conflict with https://github.com/evergreen-ci/logkeeper/pull/79 will probably be significant, but I can deal with that after it merges.

I wrote it to rely on pulling S3 credentials out of the environment. This is already how we get our [slack](https://github.com/evergreen-ci/logkeeper/blob/ed37c6d0c343b74c08cd81feb78ec22cba8931fd/senders.go#L47-L48)/[evergreen ](https://github.com/evergreen-ci/logkeeper/blob/cb88682ab734bf7b06f86d67dae87c230df17b92/units/cleanup_old_log_data.go#L26-L27) credentials. Before we deploy this we will need to make sure that the credentials are in the application's environment or the application won't start.


main/logkeeper.go
```diff
@@ -34,6 +35,7 @@ func main() {
 	dbHost := flag.String("dbhost", "localhost:27017", "host/port to connect to DB server. Comma separated.")
 	rsName := flag.String("rsName", "", "name of replica set that the DB instances belong to. "+
 		"Leave empty for stand-alone and mongos instances.")
+	localPath := flag.String("localPath", "", "local path to save data to. Omit to save data to S3.")
```
johndaniels, line 38: We probably will want to make the S3 bucket configurable. I'm guessing we think it would be simpler to make that a second parameter? I think that makes sense...

main/logkeeper.go
```diff
@@ -163,3 +169,14 @@ func gracefulShutdownForSIGTERM(ctx context.Context, servers []*http.Server, gra
...
+
+func makeBucket(localPath *string) (storage.Bucket, error) {
+	if localPath != nil {
+		return storage.NewBucket(storage.BucketOpts{
+			Location: storage.PailLocal,
+			Path:     *localPath,
+		})
+	}
+
+	return storage.NewBucket(storage.BucketOpts{Location: storage.PailS3})
```
johndaniels, line 181: Does this not require a bucket name?

makefile
```diff
@@ -121,7 +121,7 @@ $(buildDir)/output.%.coverage: .FORCE
 $(buildDir)/output.%.coverage.html:$(buildDir)/output.%.coverage
 	go tool cover -html=$< -o $@
 $(buildDir)/output.smoke.test: $(buildDir)/$(name)
-	./$< &
+	./$< --localPath $(buildDir) &
```
johndaniels, line 124: Does it make sense to use _bucketdata for this, or possibly a subfolder under buildDir? I'd be worried about writing to the build directory itself.

storage/persistence_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"os"
+	"testing"
+
+	"github.com/evergreen-ci/logkeeper/model"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUploadBuildMetadata(t *testing.T) {
+	storage := makeTestStorage(t, "")
+	defer os.RemoveAll(tempDir)
```
johndaniels, line 15: Not sure if we want to do this in this change, but we may want to make a separate cleanup function for the test storage so that it's at the same level of abstraction as "makeTestStorage" (that is we're not just removing tempDir in all of our functions).

storage/persistence_test.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	assert.NoError(t, storage.UploadBuildMetadata(context.Background(), build))
+	results, err := storage.Get(context.Background(), "5a75f537726934e4b62833ab6d5dca41/metadata.json")
+	assert.NoError(t, err)
+	contents, err := io.ReadAll(results)
+	assert.NoError(t, err)
+
+	expectedMetadata := `{"id":"5a75f537726934e4b62833ab6d5dca41","builder":"builder0","buildnum":1,"task_id":"t0"}`
+	assert.Equal(t, expectedMetadata, string(contents))
```
johndaniels, line 31: Just to confirm Go Json serialization always produces a stable key order?

storage/storage.go
```diff
@@ -1,146 +1,92 @@
...
 
-	iterator, listErr := storage.bucket.List(context, buildPrefix)
-	buildChunks := []LogChunkInfo{}
-	if listErr != nil {
-		return nil, listErr
-	}
-	for iterator.Next(context) {
-		if strings.HasSuffix(iterator.Item().Name(), "metadata.json") {
-			continue
+		return Bucket{localBucket}, nil
```
johndaniels, line 51: I need to remember this syntax.

storage/storage_model.go
```diff
@@ -0,0 +1,105 @@
...
+	var prefix string
+	if info.TestID != "" {
+		prefix = testPrefix(info.BuildID, info.TestID)
+	} else {
+		prefix = buildPrefix(info.BuildID)
+	}
+	return fmt.Sprintf("%s%d_%d_%d", prefix, info.Start.UnixNano(), info.End.UnixNano(), info.NumLines)
+}
+
+func (info *LogChunkInfo) fromKey(path string) error {
```
johndaniels, line 36: Should this be a static function called logChunkFromKey instead? Or is populating an existing object a better pattern?

storage/storage_model_test.go
```diff
@@ -0,0 +1,62 @@
+package storage
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLogChunkInfoKey(t *testing.T) {
```
johndaniels, line 10: Nice tests!

storage/storage_test.go
```diff
@@ -10,98 +10,102 @@ import (
...
 	assert.NotEqual(t, nil, results)
 
 }
 
-func TestGetTestLogLines(t *testing.T) {
-	storage := makeTestStorage(t, "../testdata/simple")
-	defer os.RemoveAll("../_bucketdata")
-	iterator, err := storage.GetTestLogLines(context.Background(), "5a75f537726934e4b62833ab6d5dca41", "62dba0159041307f697e6ccc")
-	require.NoError(t, err)
+func TestGetS3Options(t *testing.T) {
```
johndaniels, line 47: If we want to use S3 for our tests then we are definitely going to need to add a "cleanup" method that knows how to handle that.

johndaniels (changes requested):

storage/persistence_test.go
johndaniels: This is more important if we use an S3 bucket optionally in tests. We can have a "cleanupTestStorage" function that knows how to clear the temporary directory. I tried using the bucket's remove function but that only deletes files and not directories if you are using a filesystem based storage.

main/logkeeper.go
johndaniels: I see, this is configured by the s3options that also gets secrets etc.

storage/storage_model.go
johndaniels: Ah this C style initialization avoids some weirdness with returning pointers, doesn't it?

storage/persistence_test.go
ybrill: I think so ([docs](https://pkg.go.dev/encoding/json#:~:text=The%20map%20keys%20are%20sorted))

storage/storage_model.go
ybrill: It's a pattern we use a bunch. I don't know that there's an advantage besides keeping the package namespace clear.
I'm fine either way.

PR #82, opened 2022-08-08
EVG-17158: Add ability for view route to fetch from S3.

johndaniels:
This adds an "s3" build parameter to view the logs for a test
that is stored in s3, and all the infrastructure necessary to
do that.

Interestingly, the current code for fetching log lines from mongo
doesn't handle errors, so when we translate our log iterator to
the existing interface, we silently ignore errors. We may want
to surface these better.


storage/storage_model.go
```diff
@@ -103,3 +119,55 @@ func (m *buildMetadata) toJSON() ([]byte, error) {
...
+type testMetadata struct {
+	ID      string `json:"id"`
+	Name    string `json:"name"`
+	BuildID string `json:"build_id"`
+	TaskID  string `json:"task_id"`
+}
+
+func newTestMetadata(t model.Test) testMetadata {
+	return testMetadata{
+		ID:      t.Id.String(),
```
ybrill, line 136: I think we want `t.Id.Hex()` since the String representation adds some extra cruft.

storage/retrieval.go
```diff
@@ -135,3 +137,45 @@ func (storage *Bucket) GetTestLogLines(context context.Context, buildId string,
...
+
+	bytes, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return nil, errors.Wrapf(err, "reading build metadata for build '%s'", id)
+	}
+	metadata := buildMetadata{}
+	err = metadata.fromJSON(bytes)
+	if err != nil {
+		return nil, errors.Wrapf(err, "parsing build metadata for build '%s'", id)
+	}
```
ybrill, line 156: Maybe it makes sense to use a json [Decoder](https://pkg.go.dev/encoding/json#NewDecoder) since it takes care of reading/unmarshaling from a reader?

views.go
```diff
@@ -391,6 +397,70 @@ func (lk *logKeeper) viewAllLogs(w http.ResponseWriter, r *http.Request) {
...
+		defer close(logsChan)
+		for iterator.Next(r.Context()) {
+			if iterator.Err() != nil {
+				lk.logErrorf(r, "Error iterating over logs: %v", err)
+				break
+			}
+			item := iterator.Item()
+			logsChan <- &item
+		}
+	}()
```
ybrill, line 454: since this is the primary place we'll be using the return of `GetTestLogLines` could it make sense for it to return a logsChan?

views.go
```diff
@@ -391,6 +397,70 @@ func (lk *logKeeper) viewAllLogs(w http.ResponseWriter, r *http.Request) {
...
+	test, err := lk.opts.Bucket.FindTestByID(r.Context(), buildID, testID)
+	if err != nil || test == nil {
+		return nil, &apiError{Err: "view test in S3 by id: test not found", code: http.StatusNotFound}
+	}
+
+	iterator, err := lk.opts.Bucket.GetTestLogLines(r.Context(), buildID, testID)
+	if err != nil {
+		lk.logErrorf(r, "Error finding logs during test: %v", err)
+		return nil, &apiError{Err: err.Error(), code: http.StatusInternalServerError}
+	}
```
ybrill, line 440: do we want to parallelize these three requests?

views.go
```diff
@@ -391,6 +397,70 @@ func (lk *logKeeper) viewAllLogs(w http.ResponseWriter, r *http.Request) {
...
+		build:    build,
+		test:     test,
+	}
+	return &result, nil
+}
+
+func (lk *logKeeper) viewTestInS3(r *http.Request, buildID string, testID string) (*logFetchResponse, *apiError) {
+	build, err := lk.opts.Bucket.FindBuildByID(r.Context(), buildID)
+	if err != nil || build == nil {
+		return nil, &apiError{Err: "view test in S3 by id: build not found", code: http.StatusNotFound}
```
ybrill, line 428: If `err != nil` I'd think we'd want to return like a 500 or something.

views.go
```diff
@@ -391,6 +397,70 @@ func (lk *logKeeper) viewAllLogs(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (lk *logKeeper) viewTestInDatabase(r *http.Request, buildID string, testID string) (*logFetchResponse, *apiError) {
```
ybrill, line 400: as an alternative to returning a pointer to the error we could return an `error` (which is an interface) and return an `apiError` in case of error. 
The caller could then cast/test its type and pull out the relevant fields.

storage/storage_model.go
```diff
@@ -103,3 +119,55 @@ func (m *buildMetadata) toJSON() ([]byte, error) {
...
+func (m *buildMetadata) fromJSON(data []byte) error {
+	return errors.Wrap(json.Unmarshal(data, m), "unmarshalling metadata")
+}
+
+type testMetadata struct {
+	ID      string `json:"id"`
+	Name    string `json:"name"`
+	BuildID string `json:"build_id"`
+	TaskID  string `json:"task_id"`
+}
```
ybrill, line 132: Just to make sure: we're not holding on to the phase and command (from [here](https://github.com/evergreen-ci/logkeeper/blob/3de44c77ad289efb1e04246e61e03a3da37e5ef3/views.go#L171-L172)) because we never use them?

storage/retrieval.go
johndaniels: Thanks for the link to the docs!

views.go
johndaniels: Possibly... depends on if we want to add the ability to pass errors through from there. One reason the Iterator interface exists is to be able to retrieve errors using the Err() function. I'll give it a try with the channel though since we do ignore errors.

storage/storage_model.go
johndaniels: Yeah, I don't think we expose it anywhere? We don't serialize the json of our tests anywhere AFAICT?

views.go
johndaniels: I was trying to be as restrictive as possible about the type returned and keep an interface of "the thing this returns will always be ready to be immediately serialized into a response". If the Go convention is to only return error I can switch to that but this keeps the calling code simpler.

storage/storage_model.go
ybrill: I don't see it anywhere either.
We use name [here](https://github.com/evergreen-ci/logkeeper/blob/0bd05c4d927524c21ee7cc48d2e87767b783b644/views.go#L522), but I don't see anything else.

views.go
johndaniels: This is the behavior from the original code, but I think it makes sense to go ahead and update it to return the right thing.

views.go
johndaniels: Made a commit to use channels instead. LMK what you think. It's easy to revert if we hate it.

views.go
johndaniels: I parallelized them. Let me know what you think about whether it's too complicated or if there's a better way to do that.

views.go
johndaniels: This is fixed in the parallelized version.

testdata/overlapping/5a75f537726934e4b62833ab6d5dca41/metadata.json
```diff
@@ -1,9 +1,6 @@
 {
-    "_id": "5a75f537726934e4b62833ab6d5dca41",
+    "id": "5a75f537726934e4b62833ab6d5dca41",
     "builder": "builder",
     "buildnum": 157865447,
-    "name": "157865447",
     "task_id": "A task",
```
ybrill, line 5: vscode flagged the trailing comma here

storage/retrieval.go
```diff
@@ -106,7 +110,25 @@ func getFirstTestChunkAfter(allTestChunks []LogChunkInfo, target time.Time) time
 	return TimeRangeMax
 }
 
-func (storage *Bucket) GetTestLogLines(context context.Context, buildId string, testId string) (LogIterator, error) {
+func channelFromIterator(context context.Context, iterator LogIterator) chan *model.LogLineItem {
```
ybrill, line 113: could it make sense to build this into the iterator? Like iterator.Channel(ctx) could return a channel

views.go
```diff
@@ -391,6 +398,87 @@ func (lk *logKeeper) viewAllLogs(w http.ResponseWriter, r *http.Request) {
...
+	var logsChanErr error
+	go func() {
+		defer wg.Done()
+		logsChan, logsChanErr = lk.opts.Bucket.GetTestLogLines(r.Context(), buildID, testID)
+	}()
+
+	wg.Wait()
+
+	if buildErr != nil {
+		return nil, &apiError{Err: "view test in S3 by id: error fetching build", code: http.StatusInternalServerError}
```
ybrill, line 454: In the `StatusInternalServerError` cases I see why we don't want to tell the client what the error was, but does it make sense to log it? We could pass it to `lk.logErrorf`.

storage/retrieval.go
```diff
@@ -106,7 +110,25 @@ func getFirstTestChunkAfter(allTestChunks []LogChunkInfo, target time.Time) time
...
+	logsChan := make(chan *model.LogLineItem)
+
+	go func() {
```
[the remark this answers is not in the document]
johndaniels: Yeah that makes sense that it wouldn't be able to recover from a different goroutine. Kind of wild that the whole program crashes when a goroutine panics though.

storage/retrieval.go
johndaniels: I can't figure out a way to add a method to an interface, otherwise this would be ideal.

views.go
johndaniels: So you're saying we think we shouldn't return StatusInternalServerError?

storage/retrieval.go
```diff
@@ -133,5 +155,45 @@ func (storage *Bucket) GetTestLogLines(context context.Context, buildId string,
...
+
+	bytes, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return nil, errors.Wrapf(err, "reading test metadata for build: '%s' and test: '%s'", buildId, testId)
+	}
+	metadata := testMetadata{}
+	err = metadata.fromJSON(bytes)
+	if err != nil {
+		return nil, errors.Wrapf(err, "parsing test metadata for build: '%s' and test: '%s'", buildId, testId)
+	}
```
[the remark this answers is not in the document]
johndaniels: This should be fixed.

storage/retrieval.go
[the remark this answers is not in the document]
johndaniels: Added the recovery.

views.go
ybrill: I'm with you on returning `StatusInternalServerError` to the client. I was just suggesting that we log the actual error in `buildErr` to Splunk or something. This way if S3 starts returning a particular error, we'll be able to see it in Splunk.

storage/retrieval.go
```diff
@@ -106,7 +110,25 @@ func getFirstTestChunkAfter(allTestChunks []LogChunkInfo, target time.Time) time
...
-func (storage *Bucket) GetTestLogLines(context context.Context, buildId string, testId string) (LogIterator, error) {
+func channelFromIterator(context context.Context, iterator LogIterator) chan *model.LogLineItem {
+	logsChan := make(chan *model.LogLineItem)
+
+	go func() {
+		defer close(logsChan)
+		for iterator.Next(context) {
+			if iterator.Err() != nil {
+				grip.Errorf("Error iterating over logs: %v", iterator.Err())
+				break
```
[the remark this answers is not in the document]
johndaniels: It looks like it will return False for next on an unrecoverable error, so we can just defer and grab/log the errors at the end which just calls catcher.Resolve().

views.go
```diff
@@ -391,6 +398,87 @@ func (lk *logKeeper) viewAllLogs(w http.ResponseWriter, r *http.Request) {
...
+	return &result, nil
+}
+
+func (lk *logKeeper) viewTestInS3(r *http.Request, buildID string, testID string) (*logFetchResponse, *apiError) {
+	var build *model.Build
+	var buildErr error
+	wg := sync.WaitGroup{}
+
+	wg.Add(3)
+	go func() {
```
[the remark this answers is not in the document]
johndaniels: Added.

testdata/overlapping/5a75f537726934e4b62833ab6d5dca41/metadata.json
johndaniels: Fixed.

storage/retrieval.go
johndaniels: Open to suggestions. I know I could wrap it in another struct but that seems not worth it?

views.go
johndaniels: Updated these to log the error, but not return that result to the user.

johndaniels: Happy to change this again if I interpreted this wrong though.

storage/retrieval.go
ybrill: I added a commit (8c6aa24) with what I'd been thinking. 
(definitely feel free to revert it if it doesn't make sense)

views.go
ybrill: I'd been thinking more along the lines of 
```go
if buildErr != nil {
	lk.logErrorf(r, "error fetching build: %v", buildErr)
	return nil, &apiError{Err: "error fetching build", code: http.StatusInternalServerError}
if build == nil {
	return nil, &apiError{Err: fmt.Sprintf("no matching build for id '%s'", buildID), code: http.StatusNotFound}
}
```
This way let's say AWS returns an error we'll be logging the actual error.

johndaniels:
:D Ok fine I got you to do my work for me.

PR #86, opened 2022-08-09
EVG-17410: Add gzipping to logkeeper routes.

johndaniels:
This adds gzip to the major log returning routes, which should
substantially reduce their size.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/evergreen-ci/pail v0.0.0-20220705141756-9b2747c62b29
 	github.com/evergreen-ci/render v0.0.0-20141211045555-c9e0e54c798f
 	github.com/evergreen-ci/utility v0.0.0-20220404192535-d16eb64796e6
+	github.com/gorilla/handlers v1.5.1 // indirect
 	github.com/gorilla/mux v1.8.0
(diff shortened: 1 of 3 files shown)


ybrill:
I just remembered: [evergreen uses a gzip middleware](https://github.com/evergreen-ci/evergreen/blob/eef4876f31be3579762f69400853904fdf358f0c/service/service.go#L50) that gzips the response [whenever it will be accepted](https://github.com/phyber/negroni-gzip/blob/c20bb8813a75c1595b380259d248c5d9a1739bb3/gzip/gzip.go#L115-L119). Should we consider using that instead?

johndaniels:
Per conversation in Slack: Yes. Yes we should

julianedwards (approved):

ybrill:
I think we could `go mod tidy` to make the dependency direct.

ybrill:
Could it make sense to wrap everything in the handler (like in [the example in the readme](https://github.com/gorilla/handlers#example))?

johndaniels:
@ybrill we could wrap everything... most of our non-log responses are tiny so I'm not sure it's much of a win if any if we gzip them though.

ybrill:
sgtm

ybrill (approved):

PR #87, opened 2022-08-09
EVG-17153: Add ability to upload log chunks to persistence.go

johndaniels:
Finishes both EVG-17153 and EVG-17152.

### DIFF
--- a/storage/persistence_test.go
+++ b/storage/persistence_test.go
@@ -2,7 +2,9 @@ package storage
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"strings"
 	"testing"
 	"time"
 
@@ -11,6 +13,18 @@ import (
 	"github.com/stretchr/testify/require"
 	"gopkg.in/mgo.v2/bson"
 )
+
+type expectedChunk struct {
+	filename string
+	body     string
+}
+
+func newExpectedChunk(filename string, lines []string) expectedChunk {
+	return expectedChunk{
+		filename: filename,
+		body:     strings.Join(lines, ""),
+	}
+}
 
 func TestUploadBuildMetadata(t *testing.T) {
 	storage := makeTestStorage(t, "")
@@ -54,6 +68,18 @@ func TestUploadTestMetadata(t *testing.T) {
 	assert.JSONEq(t, expectedMetadata, string(contents))
 }
 
+// Checks that we stored the data in the right location and text format
+func verifyDataStorage(t *testing.T, storage Bucket, prefix string, expectedChunks []expectedChunk) {
+	for _, expectedChunk := range expectedChunks {
+		actualChunkStream, err := storage.Get(context.Background(), fmt.Sprintf("%s%s", prefix, expectedChunk.filename))
+		require.NoError(t, err)
+
+		actualChunkBody, err := io.ReadAll(actualChunkStream)
+		require.NoError(t, err)
+		assert.Equal(t, expectedChunk.body, string(actualChunkBody))
+	}
+}
+
 func TestInsertLogChunks(t *testing.T) {
 	uploadChunks := []model.LogChunk{
 		{
@@ -86,6 +112,18 @@ func TestInsertLogChunks(t *testing.T) {
 		},
 	}
 
+	expectedChunks := []expectedChunk{
+		newExpectedChunk("1000000000000000000_1000000002000000000_3", []string{
+			"  0       1000000000000line0\n",
+			"  0       1000000001000line1\n",
+			"  0       1000000002000line2\n",
+		}),
+		newExpectedChunk("1000000003000000000_1000000005000000000_3", []string{
+			"  0       1000000003000line3\n",
+			"  0       1000000004000line4\n",
+			"  0       1000000005000line5\n",
+		}),
+	}
 	expected := []model.LogLineItem{
 		{
 			LineNum:   0,
@@ -127,6 +165,8 @@ func TestInsertLogChunks(t *testing.T) {
 		err := storage.InsertLogChunks(context.Background(), buildID, "", uploadChunks)
 		require.NoError(t, err)
 
+		verifyDataStorage(t, storage, fmt.Sprintf("/builds/%s/", buildID), expectedChunks)
+
 		logsChannel, err := storage.GetAllLogLines(context.Background(), buildID)
 		require.NoError(t, err)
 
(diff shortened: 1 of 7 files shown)


johndaniels:
Note: Add non-rtt testing.

storage/persistence_test.go
```diff
@@ -51,3 +67,138 @@ func TestUploadTestMetadata(t *testing.T) {
 	expectedMetadata := `{"id":"62dba0159041307f697e6ccc","name":"test0","build_id":"5a75f537726934e4b62833ab6d5dca41","task_id":"t0"}`
 	assert.JSONEq(t, expectedMetadata, string(contents))
 }
+
+// Checks that we stored the data in the right location and text format
```
julianedwards, line 71: Nit: remove this, no need to add function documentation in tests.

views.go
```diff
@@ -267,6 +267,14 @@ func (lk *logKeeper) appendLog(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if build.S3 {
+		if err := lk.opts.Bucket.InsertLogChunks(r.Context(), build.Id, test.Id.Hex(), chunks); err != nil {
+			lk.logErrorf(r, "appending s3 logs: %v", err)
```
julianedwards, line 272: Nit: `S3`.

views.go
```diff
@@ -326,6 +334,14 @@ func (lk *logKeeper) appendGlobalLog(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if build.S3 {
+		if err := lk.opts.Bucket.InsertLogChunks(r.Context(), build.Id, "", chunks); err != nil {
+			lk.logErrorf(r, "appending s3 logs: %v", err)
```
julianedwards, line 339: Nit: `S3`.

storage/storage_model.go
```diff
@@ -68,6 +85,28 @@ func (info *LogChunkInfo) fromKey(path string) error {
 	return nil
 }
 
+func (info *LogChunkInfo) fromLogChunk(buildID string, testID string, logChunk model.LogChunk) error {
+	if len(logChunk) == 0 {
+		return errors.New("Log chunk must contain at least one line")
+	}
+	minTime := TimeRangeMax
+	maxTime := TimeRangeMin
+	for _, logLine := range logChunk {
```
julianedwards, line 94: Should we ensure that log lines are in sorted ts order, at least within a chunk?

storage/storage_model.go
```diff
@@ -68,6 +85,28 @@ func (info *LogChunkInfo) fromKey(path string) error {
 	return nil
 }
 
+func (info *LogChunkInfo) fromLogChunk(buildID string, testID string, logChunk model.LogChunk) error {
+	if len(logChunk) == 0 {
+		return errors.New("Log chunk must contain at least one line")
```
julianedwards, line 90: Nit: `log...`

storage/storage_model.go
```diff
@@ -14,6 +14,23 @@ import (
...
 
+func parseLogLineString(data string) (model.LogLineItem, error) {
+	ts, err := strconv.ParseInt(strings.TrimSpace(data[3:23]), 10, 64)
+	if err != nil {
+		return model.LogLineItem{}, err
+	}
+
+	return model.LogLineItem{
+		Timestamp: time.Unix(0, ts*1e6).UTC(),
+		// We need to Trim the newline here because logkeeper doesn't expect newlines to be included in the LogLineItem
```
julianedwards, line 25: Nit: add a period and `Logkeeper`.

storage/storage_model.go
```diff
@@ -14,6 +14,23 @@ import (
 
 const metadataFilename = "metadata.json"
 
+func parseLogLineString(data string) (model.LogLineItem, error) {
+	ts, err := strconv.ParseInt(strings.TrimSpace(data[3:23]), 10, 64)
+	if err != nil {
+		return model.LogLineItem{}, err
```
julianedwards, line 20: Wrap this error.

storage/storage_model.go
```diff
@@ -14,6 +14,23 @@ import (
...
+
+	return model.LogLineItem{
+		Timestamp: time.Unix(0, ts*1e6).UTC(),
+		// We need to Trim the newline here because logkeeper doesn't expect newlines to be included in the LogLineItem
+		Data: strings.TrimRight(data[23:], "\n"),
+	}, nil
+}
+
+func makeLogLineString(logLine model.LogLine) string {
+	return fmt.Sprintf("  0%20d%s\n", logLine.Time.UnixNano()/1e6, logLine.Msg)
```
julianedwards, line 31: We have a function [UnixMilli](https://github.com/evergreen-ci/utility/blob/main/time.go#L26) in our `utility` package that will do the conversion for you.

julianedwards (changes requested):

storage/storage_model.go
johndaniels: So the current implementation just assumes that it is, without ensuring that. I'm not inclined to do additional sorting at least in this change since it would be a significant functional change if people are sending us stuff out of order.

storage/storage_model.go
johndaniels: It's built in after 1.8.....

storage/persistence_test.go
johndaniels: I don't 100% agree with never adding function documentation in tests, but in this case it's a very "meh" and not super informative comment so I removed it.

views.go
johndaniels: Fixed

views.go
johndaniels: fixed

storage/storage_model.go
johndaniels: Fixed

storage/storage_model.go
johndaniels: Fixed

storage/storage_model.go
johndaniels: Wrapped

storage/storage_model.go
johndaniels: Fixed. Thanks for the heads up!

storage/storage_model.go
julianedwards: If we are assuming then do we need to find the min and the max? Can't we just grab the first and last timestamp from the chunk?

julianedwards (approved):

PR #91, opened 2022-08-10
EVG-17150 upload test metadata

ybrill:
[EVG-17150](https://jira.mongodb.org/browse/EVG-17150)

Write test metadata to S3. Mostly the same as creating a build.

I added a field to the build document in the db that will indicate if we're duplicating to S3. If we are then all subsequent operations go to S3 too. This strategy is just for testing. In the goal state we'll stop doing this and just write everything to S3 (instead of the db).

### DIFF
--- a/views.go
+++ b/views.go
@@ -124,6 +124,7 @@ func (lk *logKeeper) createBuild(w http.ResponseWriter, r *http.Request) {
 		Name:     fmt.Sprintf("%v #%v", buildParameters.Builder, buildParameters.BuildNum),
 		Started:  time.Now(),
 		Info:     model.BuildInfo{TaskID: buildParameters.TaskId},
+		S3:       buildParameters.S3,
 	}
 	if err = newBuild.Insert(); err != nil {
 		lk.logErrorf(r, "Error inserting build object: %v", err)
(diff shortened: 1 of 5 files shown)


views.go
```diff
@@ -197,6 +198,14 @@ func (lk *logKeeper) createTest(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if build.S3 {
```
johndaniels, line 201: Isn't the intent that if the build was stored in S3, it wouldn't have an entry in the database? This change doesn't handle that case... not sure if that's something that should be included here or in a later change.

johndaniels (approved):

johndaniels:
This is fine to merge and we can figure out the case of returning when the build isn't in the database

views.go
ybrill: My thinking is that there are three stages:
1. Now. Production data is stored in the db. If we specify we want something to go to S3 (for testing) then it's duplicated there.
2. When we're writing some percentage of production data to S3. More specifically, we'll write all the data for some percentage of _builds_ to S3. Here too everything is duplicated between the db and S3.
3. When production data all goes to S3. At this point data will not be written to the db at all.

In stages 1 and 2 we'll have this flag in the build document in the db so we know where to send the rest of a build. When we transition to stage 3 we know that _everything_ we're writing goes to S3 so we won't need this flag anymore.

johndaniels: Ah that makes a lot of sense. Thanks.

PR #93, opened 2022-08-11
EVG-17467 marshal/unmarshal LogLine

ybrill:
[EVG-17467](https://jira.mongodb.org/browse/EVG-17467)

Since https://github.com/evergreen-ci/logkeeper/commit/cb88682ab734bf7b06f86d67dae87c230df17b92 a LogLine was getting marshalled to/from BSON as an object. This isn't compatible with existing logs.

Implement the [Getter](https://pkg.go.dev/gopkg.in/mgo.v2@v2.0.0-20190816093944-a6b53ec6cb22/bson#Getter) and [Setter](https://pkg.go.dev/gopkg.in/mgo.v2@v2.0.0-20190816093944-a6b53ec6cb22/bson#Setter) interfaces so the driver will marshal/unmarshal in the original format. Test `GetBSON` by verifying that LogLine is marshalled into the db is as expected and test `SetBSON` by verifying that a LogLine is unmarshalled from the db.

### DIFF
--- a/model/log.go
+++ b/model/log.go
@@ -287,7 +287,7 @@ func (ll *LogLine) SetBSON(raw bson.Raw) error {
 
 	time, ok := line[0].(time.Time)
 	if !ok {
-		return errors.Errorf("timestamp was of unexpected type '%T'", line[0])
+		return errors.Errorf("timestamp was of unexpected type %T", line[0])
 	}
 
 	msg, ok := line[1].(string)
--- a/model/log.go
+++ b/model/log.go
@@ -3,6 +3,7 @@ package model
 import (
 	"encoding/json"
 	"math"
+	"reflect"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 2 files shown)


model/log.go
```diff
@@ -262,6 +263,43 @@ func (ll *LogLine) UnmarshalJSON(data []byte) error {
...
+			Type: reflect.TypeOf(line),
+		}
+	}
+	if len(line) < 2 {
+		return errors.Errorf("line was of unexpected length %d", len(line))
+	}
+
+	time, ok := line[0].(time.Time)
+	if !ok {
+		return errors.Errorf("timestamp was of unexpected type '%T'", line[0])
```
julianedwards, line 290: Nit: No need to put single quotes around %T since those are predefined values (we usually use single quotes when it is a string that may have special characters such as spaces that make it hard to delimit).

julianedwards (approved):
LGTM % nit

model/log.go
ybrill: done

PR #95, opened 2022-08-12
EVG-17493: Generate build keys from builder and buildNum.

johndaniels:
This will allow us to constant time lookup builds based on
builder and buildNum without having to add an additional S3
hierarchy.

### DIFF
--- a/model/build.go
+++ b/model/build.go
@@ -1,12 +1,10 @@
 package model
 
 import (
-	"bytes"
 	"context"
 	"crypto/md5"
-	"encoding/binary"
 	"encoding/hex"
-	"strconv"
+	"encoding/json"
 	"time"
 
 	"github.com/evergreen-ci/logkeeper/db"
@@ -155,29 +153,22 @@ func RemoveBuild(buildID string) error {
 	return errors.Wrap(db.C(BuildsCollection).RemoveId(buildID), "deleting build record")
 }
 
-const (
-	builderFieldNum  = 1
-	buildNumFieldNum = 2
-)
-
-func makeBinaryRepresentation(builder string, buildNum int) []byte {
-	var buf bytes.Buffer
-	builderBytes := []byte(builder)
-	buildNumBytes := []byte(strconv.Itoa(buildNum))
-	binary.Write(&buf, binary.BigEndian, uint32(builderFieldNum))
-	binary.Write(&buf, binary.BigEndian, uint32(len(builderBytes)))
-	buf.Write(builderBytes)
-	binary.Write(&buf, binary.BigEndian, uint32(buildNumFieldNum))
-	binary.Write(&buf, binary.BigEndian, uint32(len(buildNumBytes)))
-	buf.Write(buildNumBytes)
-	return buf.Bytes()
-}
-
 // Generates a new build ID based on the hash of builder and buildNum
 func NewBuildId(builder string, buildNum int) (string, error) {
 	hasher := md5.New()
 
-	if _, err := hasher.Write(makeBinaryRepresentation(builder, buildNum)); err != nil {
+	// This depends on the fact that Go's json implementation sorts json keys
+	// lexicographically for maps, which ensures consistent encoding
+	jsonMap := make(map[string]interface{})
+	jsonMap["builder"] = builder
+	jsonMap["buildNum"] = buildNum
+	hashstring, err := json.Marshal(jsonMap)
+
+	if err != nil {
+		return "", errors.Wrap(err, "generating json to hash for build key")
+	}
+
+	if _, err := hasher.Write(hashstring); err != nil {
 		return "", errors.Wrap(err, "hashing json for build key")
 	}
 
--- a/model/build.go
+++ b/model/build.go
@@ -168,7 +168,7 @@ func NewBuildId(builder string, buildNum int) (string, error) {
 		return "", errors.Wrap(err, "generating json to hash for build key")
 	}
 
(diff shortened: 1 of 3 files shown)


model/build.go
```diff
@@ -149,3 +152,25 @@ func RemoveBuild(buildID string) error {
...
+	// This depends on the fact that Go's json implementation sorts json keys
+	// lexicographically for maps, which ensures consistent encoding
+	jsonMap := make(map[string]interface{})
+	jsonMap["builder"] = builder
+	jsonMap["buildNum"] = buildNum
+	hashstring, err := json.Marshal(jsonMap)
+
+	if err != nil {
+		return "", errors.Wrap(err, "generating json to hash for build key")
+	}
```
ybrill, line 169: Given the result is going to be hashed, why is it desirable to marshal to JSON first?
Like couldn't we just concatenate the builder and buildnum? Or `fmt.Sprintf("%s-%d", builder, buildNum)`

views.go
```diff
@@ -110,13 +108,12 @@ func (lk *logKeeper) createBuild(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	hasher := md5.New()
-	if _, err = hasher.Write([]byte(bson.NewObjectId().Hex())); err != nil {
+	newBuildId, err := model.NewBuildId(buildParameters.Builder, buildParameters.BuildNum)
```
ybrill, line 111: should we also check, in the case where `buildParameters.S3 == true` , that the build doesn't exist in S3?

model/build.go
johndaniels: Ugh, I spent a fair bit of time going mentally back and forth on this.

The big reason is collisions and forward compatibility. In order to avoid collisions, the function needs to be injective (for every element in the output, there is only one input that will give that output). I thought about doing it with fmt.Sprintf, and there _shouldn't_ be the possibility of collision since buildNum is guaranteed not to have a hyphen (so you can't have ("asdf-123", 456) colliding with ("asdf", "123-456")), but the advantage of JSON is that it's already an encoding with escaping rules and no collisions already in place (I think we are all convinced that json strings can always be unambiguously decoded to a consistent set of key/value pairs). It also means that if we wanted to add a 3rd, optional value that'd be really easy. If for example we added another string, we don't have to figure out a way to unambiguously encode that in an injective way that doesn't collide with the existing encoding of two elements.

Unfortunately, JSON has its own problems, the biggest of which is concerns is that the encoding is not, strictly speaking, actually a function, primarily due to key ordering not being guaranteed to be consistent. Go does sort map keys when outputting to JSON however, so key ordering shouldn't be an issue. I'm not super attached to doing it this way, but I think it's slightly cleaner and safer?

Probably the "best" option is to specify an unambiguous binary encoding, which would probably look like:
FIELDNUM (big endian 4 bytes) FIELDLEN (big endian 4 bytes) TEXT

But that seemed scarier than just using JSON.

johndaniels: One could also do a text based encoding that's not JSON and have it be unambiguous as well, but that requires an unambiguous text grammar. One easy way is finding a separator that never occurs in your data (I thought about using the null byte). Another would be to do something similar with "FIELD_NUM FIELD_LEN\n" followed by exactly FIELD_LEN bytes of text. Another would be to pick a separator and an escape sequence and make sure you escape anything that would be a separator (and also that you escape anything that is the escape sequence followed by a separator).

johndaniels: I think my second favorite option was probably the binary encoding.

views.go
johndaniels: I think no? If we've made it past the code that checks if it's a dupe in the database, then we've already written something to the DB. I think for our sanity it's probably best to just overwrite what's in S3. 

Like long term we want to check what's in S3, but if we find out that there is in fact a dup in S3 that wasn't in the DB I feel like there's not a lot we can do? I could log an error I suppose?

johndaniels: Maybe logging an error is a good idea either way. Do you think we should write the data or just log an error and refuse to write to S3?

model/build.go
ybrill: Ah, thank you for the edifying explanation.

views.go
ybrill: Ah, I missed that this is just going to be for the current stage (where we still write everything to the db). It seems we don't need to worry about it since we're guaranteed that we'll never get past the db dupe check if it's already been created.

ybrill (approved):

model/build.go
```diff
@@ -149,3 +154,32 @@ func RemoveBuild(buildID string) error {
...
+const (
+	builderFieldNum  = 1
+	buildNumFieldNum = 2
+)
+
+func makeBinaryRepresentation(builder string, buildNum int) []byte {
+	var buf bytes.Buffer
+	builderBytes := []byte(builder)
+	buildNumBytes := []byte(strconv.Itoa(buildNum))
+	binary.Write(&buf, binary.BigEndian, uint32(builderFieldNum))
```
ybrill, line 167: Why doesn't it work to use the buffer's Write method for this?

johndaniels: Because this is writing the integer as its binary representation (in BigEndian, because I had to pick one and that's what network stuff uses), and from what I can tell using the "binary" methods is the only way to get a byte array binary representation out of an int. 

See https://stackoverflow.com/questions/16888357/convert-an-integer-to-a-byte-array for example.

I don't actually know if we want to do it this way but I figured I'd write the code and see what you want in case it's too weird.

ybrill: Oh I get it. The Write method writes a slice of bytes

PR #98, opened 2022-08-16
EVG-17160: Add featureswitch code for writing builds.

johndaniels:
This will allow us to set a percentage of builds being written to
S3. I opted to make this a hash of the build ID rather than random
for a few reasons:

1. When we use this for ramping up fetching from S3, it will ensure
that users won't alternate between code paths for fetching the same
build. This should avoid weird bugs for clients and aid in
debugging.
2. It will make debugging in general less based on chance. If you
put the same build name and id to the same configuration, you'll
hit the same code path.

### DIFF
--- a/featureswitch/featureswitch.go
+++ b/featureswitch/featureswitch.go
@@ -14,9 +14,13 @@ import (
 const s3WriteFeatureSwitch = "LK_WRITE_S3_FEATURE_SWITCH"
 
 func hashToFloat(hash []byte) float64 {
+	// Use the first 4 bytes of the hash to construct a Uint32, since
+	// that is the size of a uint32 and the hash is longer (16 bytes).
 	value := binary.BigEndian.Uint32(hash[:4])
 	// The plus one ensures that we return a number in the range 0 to 1,
-	// inclusive of 0 and exclusive of 1.
+	// inclusive of 0 and exclusive of 1. The math in the denominator is
+	// untyped and should just be cast to a float64, so we shouldn't
+	// have to worry about overflow.
 	return float64(value) / (math.MaxUint32 + 1)
 }
 
@@ -44,6 +48,9 @@ func matchesFeatureForHash(featureSwitch string, hash []byte) bool {
 
 func matchesFeatureForString(featureSwitch string, data string) bool {
 	hasher := md5.New()
+	// Write both the featureswitch name and data into our hasher.
+	// This is  important so that we can use the same key with a
+	// different featureswitch to get a different value from the hash.
 	hasher.Write([]byte(featureSwitch))
 	hasher.Write([]byte(data))
 	hash := hasher.Sum(nil)
--- a/featureswitch/featureswitch.go
+++ b/featureswitch/featureswitch.go
@@ -1,0 +1,55 @@
+package featureswitch
+
+import (
+	"crypto/md5"
+	"encoding/binary"
+	"fmt"
+	"math"
+	"os"
+	"strconv"
+
+	"github.com/mongodb/grip"
+)
+
+const s3WriteFeatureSwitch = "LK_WRITE_S3_FEATURE_SWITCH"
+
+func hashToFloat(hash []byte) float64 {
+	value := binary.BigEndian.Uint32(hash[:4])
+	// The plus one ensures that we return a number in the range 0 to 1,
+	// inclusive of 0 and exclusive of 1.
+	return float64(value) / (math.MaxUint32 + 1)
+}
+
+func getThreshold(featureSwitch string) float64 {
+	value := os.Getenv(featureSwitch)
+	floatValue, err := strconv.ParseFloat(value, 64)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


featureswitch/featureswitch.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/mongodb/grip"
+)
+
+const s3WriteFeatureSwitch = "LK_WRITE_S3_FEATURE_SWITCH"
+
+func hashToFloat(hash []byte) float64 {
+	value := binary.BigEndian.Uint32(hash[:4])
+	// The plus one ensures that we return a number in the range 0 to 1,
+	// inclusive of 0 and exclusive of 1.
+	return float64(value) / (math.MaxUint32 + 1)
```
ybrill, line 20: Is this reliant on running on the size of an `int` being larger than 32 bits so it doesn't overflow?

ybrill (commented):
Could an alternative be to use the buildID as a seed for a rng?

featureswitch/featureswitch.go
```diff
@@ -0,0 +1,55 @@
...
+	"os"
+	"strconv"
+
+	"github.com/mongodb/grip"
+)
+
+const s3WriteFeatureSwitch = "LK_WRITE_S3_FEATURE_SWITCH"
+
+func hashToFloat(hash []byte) float64 {
+	value := binary.BigEndian.Uint32(hash[:4])
```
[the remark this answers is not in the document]
johndaniels: Correct

featureswitch/featureswitch.go
```diff
@@ -0,0 +1,55 @@
...
+	threshold := getThreshold(featureSwitch)
+	// value is guaranteed to be in the range [0, 1), which means that
+	// '<' will always return false for a threshold of 0 and always return
+	// true for a threshold of 1.0.
+	return value < threshold
+}
+
+func matchesFeatureForString(featureSwitch string, data string) bool {
+	hasher := md5.New()
+	hasher.Write([]byte(featureSwitch))
```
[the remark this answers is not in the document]
johndaniels: Because we want the result to be different depending on which feature flag you are looking up. I mostly stole this from A/B testing, where the same user id should get a different result for each test, but I think the same applies to feature flags with the same input key.

johndaniels:
I like the idea of seeding an Rng conceptually, but I'm not sure it makes the code any simpler, mainly because the seed itself needs to be an int64 which probably still requires an additional hash to make sure the entire buildID is incorporated into the seed?

featureswitch/featureswitch.go
johndaniels: It's an untyped numeric constant so it doesn't overflow. This is also captured in the tests.

featureswitch/featureswitch.go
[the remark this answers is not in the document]
johndaniels: Maybe it's too clever. I wanted to use a uint32 instead of a uint64 so that it would be exactly representable by the float64 elsewhere.

featureswitch/featureswitch.go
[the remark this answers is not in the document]
johndaniels: I can add a comment to that effect if it makes it clearer?

featureswitch/featureswitch.go
johndaniels: Basically what I _think_ happens is that it's converted to a float64 for the purposes of the division, and should be exactly representable, since it is only 33 bits.

johndaniels:
I guess it would mainly avoid a little bit of the math weirdness in hashToFloat. Thoughts?

featureswitch/featureswitch.go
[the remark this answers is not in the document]
ybrill: I get it now. Sure!

featureswitch/featureswitch.go
ybrill: How do the tests prove this out?

ybrill:
> math weirdness

Either way is good by me

featureswitch/featureswitch.go
johndaniels: I convert it to a float64 to ensure that it uses float division and not integer division (which would always give us 1 / 0).

The tests demonstrate that given 0xFFFFFFFF (max uint) we output a number very close to 1, as opposed to overflowing in some fashion and giving an odd result (+Inf seems like the most likely, since overflowing uint should go to 0 and floating point division by zero should return +Inf).

johndaniels: That overflow happens because it needs to store in variable, and it actually happens at compile time from what I can tell, since Go can tell that the constant value exceeds what it can store in an int64, which I'm guessing is the largest int type it will consider?

I an also make sure that we cast the denominator immediately to a float64 if that seems less iffy.

ybrill: Got it.
Either way is good by me

ybrill (approved):

PR #102, opened 2022-08-19
EVG-17552: Add feature switch for reading from S3.

johndaniels:
This adds a featureswitch for setting a percentage to read from
from S3, as well as the ability to fall back to reading
from the database.

This also fixes some bugs with how the view functions handled pail responses. Pail returns an error for not found, which we were not detecting properly and returning an internal server error in the event of not found.


views.go
```diff
@@ -619,20 +626,28 @@ func (lk *logKeeper) viewBucketLogs(r *http.Request, buildID string, testID stri
...
+		if shouldFallBack {
+			if testID == "" {
+				return lk.viewAllDBLogs(r, buildID)
+			} else {
+				return lk.viewDBTestLogs(r, buildID, testID)
+			}
+		} else {
+			return nil, &apiError{Err: "finding build", code: http.StatusNotFound}
+		}
+	}
```
ybrill, line 639: could it be advantageous to trigger the fallback as soon as the build goroutine returns? 
This way we wouldn't need to wait on the slowest of the requests.

views.go
```diff
@@ -434,13 +436,18 @@ func (lk *logKeeper) viewBucketBuild(r *http.Request, buildID string) (*model.Bu
 	}()
 	wg.Wait()
 
+	if pail.IsKeyNotFoundError(buildErr) {
+		if shouldFallBack {
```
ybrill, line 440: My understanding is that at the time we'll be turning this on all builds will be in the db and a subset of builds will be in S3. 
If so, could it make sense that when the feature flag indicates that we should read from S3 we 1) check the db to make sure it's in S3 and if not 2) fallback to the db. This way requests for db logs won't need to wait on the S3 responses (assuming db queries are in-fact faster)

views.go
```diff
@@ -434,13 +436,18 @@ func (lk *logKeeper) viewBucketBuild(r *http.Request, buildID string) (*model.Bu
...
+			return lk.viewDBBuild(r, buildID)
+		} else {
+			return nil, nil, &apiError{Err: "finding build", code: http.StatusNotFound}
+		}
+	}
```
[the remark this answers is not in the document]
johndaniels: I thought about it, but this avoids deeper nesting and IsKeyNotFoundError returns false when buildError is nil. Let me know if you think it reads to unclear though.

views.go
johndaniels: I moved the build fetch into the main thread because that seemed clearer/cleaner than having separate goroutine wait groups.

ybrill: Nice!
Could it make sense to pass a derived context to the other requests and cancel it when in the `pail.IsKeyNotFoundError(buildErr)` case?

johndaniels: I hope I did this right.

ybrill (approved):

PR #103, opened 2022-08-19
EVG-17531 test window

ybrill:
[EVG-17531](https://jira.mongodb.org/browse/EVG-17531)

We noticed a discrepancy between how we determined the bounds of global logs for builds in the db vs. logs in S3. For builds in the db the window is from [the creation time of this test and that of the next test](https://github.com/evergreen-ci/logkeeper/blob/a6a0c32501aa4b85a14c902ca02427a1cb588cc5/model/test.go#L182-L194). For S3 builds we're looking at the first timestamp of a this test's logs and the first timestamp of the next test's logs. This could matter:
1. when a test is created but doesn't get any logs. With the db logic the previous test's global logs ended when the next test is created while in the S3 logic the previous test's global logs don't end until a subsequent test logs something.
2. if there's a break in time from when a test is created and when it begins to log this will cause the beginning of the global logs for this test to be delayed and the end of the previous test's global logs to be extended.

This PR finds the test execution window as the time between the creation of this test and the next test. The creation time is encoded in the test ID after the fashion of an ObjectId. This is implemented by getting a new objectId from the driver and replacing the timestamp portion (the first 4 bytes) with an 8 byte timestamp with nanosecond precision. In the db, new tests will have a string type id while old tests will continue to have an ObjectId type id. We can differentiate them in the application (like if a link gives us a string) by their length (ObjectIds are 24 characters long and TestIDs are 32). SetBSON and GetBSON are implemented so we get a TestID in the application no matter what type it is in the db and so we find the correct ID in the db.

If a build already exists in S3 (currently just in staging) then we might confuse the start times of tests since the ObjectIds they have only have second precision.

Existing tests (and logs referencing their IDs) in the db will continue to have their ObjectId and the application will know how to find them. Once all the references to ObjectIDs TTL (in 120 days or when we turn off the db, whichever is sooner) we can remove the special casing (SetBSON and GetBSON) and just treat the ID as a TestID (which is a string in the db).

I also noticed the units package was missing in CI so I added it.


ybrill:
An alternative I considered was using the S3 timestamp on the metadata file, but it also only has second precision. 😞

johndaniels:
Why is the test key upper case in the testdata?

storage/retrieval.go
```diff
@@ -133,69 +153,81 @@ func (b *Bucket) getLogChunks(ctx context.Context, buildID string) ([]LogChunkIn
 		}
 	}
 
-	if err := iter.Err(); err != nil {
```
johndaniels, line 136: It looks like we stopped checking for err on iter here. Or did this move somewhere I'm missing?

storage/retrieval.go
```diff
@@ -133,69 +153,81 @@ func (b *Bucket) getLogChunks(ctx context.Context, buildID string) ([]LogChunkIn
...
+	})
+
+	return testIDs, nil
+}
+
+// testExecutionWindow returns the TimeRange from the creation of this test to the creation
+// of the next test. If the testID isn't found the returned TimeRange is unbounded.
+// If there is no later test then the end time is TimeRangeMax.
+func testExecutionWindow(allTestIDs []model.TestID, testID string) TimeRange {
+	tr := AllTime
```
johndaniels, line 191: This copies by value right? Because we mutate it later and that would be bad to mutate a global. We probably want to make TimeRange immutable at some point...

storage/retrieval.go
```diff
@@ -133,69 +153,81 @@ func (b *Bucket) getLogChunks(ctx context.Context, buildID string) ([]LogChunkIn
...
+
+	sort.Slice(testIDs, func(i, j int) bool {
+		return testIDs[i].Timestamp().Before(testIDs[j].Timestamp())
+	})
+
+	return testIDs, nil
+}
+
+// testExecutionWindow returns the TimeRange from the creation of this test to the creation
+// of the next test. If the testID isn't found the returned TimeRange is unbounded.
```
johndaniels, line 188: Should it be an error if testID isn't found?

johndaniels: Like based on my reading of the usage that should never happen.

johndaniels: The old logic had to find a TestChunk within the testid and that could fail but I don't think we can ever just not have the test metadata in the list.

johndaniels (changes requested):

johndaniels:
We should probably also add a feature switch in the NewTestId function so that we can deploy and then separately turn on actually writing the new IDs.

model/test.go
```diff
@@ -111,6 +111,69 @@ func (t *TestID) SetBSON(raw bson.Raw) error {
...
+	}
+
+	return string(*t), nil
+}
+
+// SetBSON implements the bson.Setter interface.
+// When a testIDAlias is unmarshalled from BSON the driver will call this function to
+// unmarshal into the TestID.
+func (t *testIDAlias) SetBSON(raw bson.Raw) error {
+	var id interface{}
```
johndaniels: I think you can avoid the duplication here by just converting to the TestID if not nil and deferring to that type's functions. See https://go.dev/play/p/DiDmg5YoIn8

ybrill: done

storage/retrieval.go
ybrill: got it. 
I fixed it to differentiate between the case where the testID isn't provided (like when we're fetching all the build's logs) and when a non-existent ID is provided.

storage/retrieval.go
ybrill: This got dropped.
I added it back.

storage/retrieval.go
ybrill: I think it does copy by value.
What do you mean to make a `TimeRange` immutable?

johndaniels: I don't think we need to worry about this now, but basically making the fields private so you can't change them and making public accessors. That way you don't have to worry about the whole "can someone mutate this pointer?"

ybrill:
> Why is the test key upper case in the testdata?

fixed

ybrill:
> add a feature switch in the NewTestId function

done

johndaniels (approved):

johndaniels:
Will check in tomorrow on merging/deploying this.

PR #104, opened 2022-08-19
Add ability to set and clear feature switches during tests.

johndaniels:


featureswitch/featureswitch.go
```diff
@@ -57,6 +57,22 @@ func matchesFeatureForString(featureSwitch string, data string) bool {
 	return matchesFeatureForHash(featureSwitch, hash)
 }
 
+func setFeatureSwitchLevel(featureSwitch string, level float64) func() {
+	oldValue, wasSet := os.LookupEnv(featureSwitch)
+	os.Setenv(featureSwitch, strconv.FormatFloat(level, 'f', 4, 64))
```
ybrill, line 62: Is it better to use `strconv.FormatFloat` than `fmt.Sprintf("%f")`?

featureswitch/featureswitch.go
```diff
@@ -57,6 +57,22 @@ func matchesFeatureForString(featureSwitch string, data string) bool {
 	return matchesFeatureForHash(featureSwitch, hash)
 }
 
+func setFeatureSwitchLevel(featureSwitch string, level float64) func() {
```
ybrill, line 60: I don't understand how this should be used. Is the idea that if I want to temporarily set the level from within my code I would call this and defer the function it returns?

johndaniels: Yeah that was the intent. You would use it within a test and defer the function as a "cleanup" function to get your state back to where it was. 

I could add docs to that effect.

johndaniels: Basically the issue I'm trying to solve is that I want to be able to set feature switches in tests outside of the featureswitch package, but I don't want to expose the internals of setting/unsetting the environment variables. Another option would potentially be to use a struct to manage feature switches, and replace that global struct for tests instead similar to what we do for "Environment" for Evergreen. Genuinely curious about your thoughts.

ybrill: Got it. 
The two solutions sound about the same to me. Like, the difference is just whether we use the environment for global state or a global variable.

featureswitch/featureswitch.go
johndaniels: No, I think Sprintf is honestly better.

PR #108, opened 2022-08-31
EVG-17602, EVG-17621: Parallelize and independently write to S3

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-17602, https://jira.mongodb.org/browse/EVG-17621

I made the writes to S3 as independent as possible but we still rely on the `s3` field from the build model when uploading logs and test metadata. This isn't a big deal, though, because when we are ready to remove the DB code we can just ignore it since all writes will go to S3.

I tested that this works in staging.

### DIFF
--- a/storage/persistence.go
+++ b/storage/persistence.go
@@ -32,8 +32,10 @@ func (b *Bucket) UploadTestMetadata(ctx context.Context, test model.Test) error 
 	return errors.Wrapf(b.Put(ctx, metadata.key(), bytes.NewReader(json)), "putting metadata for test '%s'", test.Id)
 }
 
-// InsertLogChunks uploads a new chunk of logs for a given build and (optional)
-// test to the offline blob bucket storage.
+// InsertLogChunks uploads a new chunk of logs for a given build or test to the
+// offline blob bucket storage. If a test ID is not empty, the logs are
+// appended to the test for the given build, otherwise the logs are appended to
+// the top-level build. A build ID is required in both cases.
 func (b *Bucket) InsertLogChunks(ctx context.Context, buildID string, testID string, chunks []model.LogChunk) error {
 	for _, chunk := range chunks {
 		if len(chunk) == 0 {
@@ -65,29 +67,4 @@ func (b *Bucket) InsertLogChunks(ctx context.Context, buildID string, testID str
 	}
 
 	return nil
-}
-
-// CheckMetadata returns whether the metadata file exists for the given build
-// and (optional) test.
-func (b *Bucket) CheckMetadata(ctx context.Context, buildID string, testID string) (bool, error) {
-	var key string
-	if testID == "" {
-		key = metadataKeyForBuild(buildID)
-	} else {
-		key = metadataKeyForTest(buildID, testID)
-	}
-
-	iter, err := b.List(ctx, key)
-	if err != nil {
-		return false, errors.Wrap(err, "listing metadata files")
-	}
-
-	for iter.Next(ctx) {
-		// We set the prefix to the entire metadata filename, so
-		// finding a single file in the bucket with the given prefix
-		// suffices.
-		return true, nil
-	}
-
-	return false, errors.Wrap(iter.Err(), "iterating over metadata files")
 }
--- a/storage/persistence.go
+++ b/storage/persistence.go
@@ -33,7 +33,7 @@ func (b *Bucket) UploadTestMetadata(ctx context.Context, test model.Test) error 
 }
 
 // InsertLogChunks uploads a new chunk of logs for a given build or test to the
-// offline blob bucket storage. If a test ID is not empty, the logs are
+// offline blob bucket storage. If the test ID is not empty, the logs are
 // appended to the test for the given build, otherwise the logs are appended to
 // the top-level build. A build ID is required in both cases.
 func (b *Bucket) InsertLogChunks(ctx context.Context, buildID string, testID string, chunks []model.LogChunk) error {
--- a/storage/persistence.go
+++ b/storage/persistence.go
@@ -8,6 +8,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+// UploadBuildMetadata uploads metadata for a new build to the offline blob
+// bucket storage.
 func (b *Bucket) UploadBuildMetadata(ctx context.Context, build model.Build) error {
 	metadata := newBuildMetadata(build)
 	json, err := metadata.toJSON()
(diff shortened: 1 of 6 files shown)


ybrill:
Did it end up to not be worthwhile to do the S3 and DB writes in parallel?
If they aren't parallel does it matter which is first?

julianedwards:
> Did it end up to not be worthwhile to do the S3 and DB writes in parallel? If they aren't parallel does it matter which is first?

It didn't seem worthwhile since we rely on the DB look ups to fetch metadata for now. I think it is still worth doing the S3 writes before the DB writes though (so those are going to be the slowest operations).

ybrill:
But I mean, couldn't we do whatever the plan will be for the post-DB era to avoid relying on the db lookups?

julianedwards:
I created https://jira.mongodb.org/browse/EVG-17621 for that work since it isn't as straightforward.

ybrill:
Ah, I understand now.
So about
> doing the S3 writes before the DB writes though (since those are going to be the slowest operations).

If the end result of the write is going to be a 500 error because of the db write taking forever why is it important to make sure the S3 write succeeds?

julianedwards:
> If the end result of the write is going to be a 500 error because of the db write taking forever why is it important to make sure the S3 write succeeds?

The DB writes don't actually use the context (old driver). So when there is high traffic and the DB slows down, the writes take much longer and by the time we get to the S3 write (which uses the context) it hits the minute timeout. We can assume that the S3 writes will be fast. If we want to be extra paranoid we can parallelize the DB and S3 writes, but it will still have to come after the DB lookups until after [EVG-17621](https://jira.mongodb.org/browse/EVG-17621).

storage/persistence.go
```diff
@@ -18,6 +20,8 @@ func (b *Bucket) UploadBuildMetadata(ctx context.Context, build model.Build) err
 	return errors.Wrapf(b.Put(ctx, metadata.key(), bytes.NewReader(json)), "putting metadata for build '%s'", build.Id)
 }
 
+// UploadBuildMetadata uploads metadata for a new test to the offline blob
```
johndaniels, line 23: what do we mean by "offline" blob bucket storage?

views.go
```diff
@@ -201,43 +233,45 @@ func (lk *logKeeper) createTest(w http.ResponseWriter, r *http.Request) {
...
+			lk.logErrorf(r, "checking metadata in build '%s': %v", buildID, err)
+			lk.render.WriteJSON(w, http.StatusInternalServerError, apiError{Err: "finding build"})
+			return
+		}
+		if !exists {
+			lk.render.WriteJSON(w, http.StatusInternalServerError, apiError{Err: "build not found"})
+			return
+		}
+	}
+	newTest.BuildName = build.Name
```
johndaniels, line 248: Can't we just move the creation of newTest down one line so that we can continue to use the object literal syntax to set BuildName?

views.go
```diff
@@ -201,43 +233,45 @@ func (lk *logKeeper) createTest(w http.ResponseWriter, r *http.Request) {
 		lk.render.WriteJSON(w, http.StatusNotFound, apiError{Err: "build not found"})
 		return
 	}
+	if build.S3 {
+		exists, err := lk.opts.Bucket.CheckMetadata(r.Context(), buildID, "")
```
johndaniels, line 237: This explicitly makes the behavior that if a build has "s3=true" but doesn't have the appropriate build metadata, we avoid doing the S3 upload/creation _and_ the DB one as well. Is that the intent here? One could also put this into the S3 parallel goroutine but it sounds like we want to just fail entirely at this point if the metadata isn't there?

storage/persistence.go
```diff
@@ -59,3 +66,28 @@ func (b *Bucket) InsertLogChunks(ctx context.Context, buildID string, testID str
 
 	return nil
 }
+
+// CheckMetadata returns whether the metadata file exists for the given build
+// and (optional) test.
```
johndaniels, line 71: This comment implies it checks both, when it actually checks the test metadata specifically if the testID is supplied.

storage/persistence.go
```diff
@@ -59,3 +66,28 @@ func (b *Bucket) InsertLogChunks(ctx context.Context, buildID string, testID str
...
+func (b *Bucket) CheckMetadata(ctx context.Context, buildID string, testID string) (bool, error) {
```
johndaniels, line 72: I know it's hard to test views.go, but we should probably at least have a small test for this function.

johndaniels (changes requested):

storage/persistence.go
julianedwards: Basically that this is not on the app or the application DB.

views.go
julianedwards: This code is to be deleted once we remove the DB writes, so I just reorganized it so it is simpler to remove it once we are ready. I am not opposed to moving it back though if you feel strongly.

views.go
julianedwards: My assumption was that we would want to fail here if `s3=true` and the build does not exist in S3 since it would be invalid to have a test without a build. I did not put it in the goroutine because the code starts to get more messy than it already is since in some cases we want to return a 404 and others a 500.

storage/persistence.go
julianedwards: Good catch, added tests!

julianedwards:
I moved `CheckMetadata` to `storage/retrieval.go` since it seemed like a more appropriate place. I also cleaned up some of the test code and added tests for `CheckMetadata`.

storage/persistence.go
julianedwards: Done.

views.go
johndaniels: Ok. Wanted to confirm

views.go
johndaniels: Ah, I see, because we plan on deleting buildname. That makes sense.

storage/storage_test.go
```diff
@@ -10,17 +10,10 @@ import (
...
 func makeTestStorage(t *testing.T, initDir string) Bucket {
-	err := os.RemoveAll(tempDir)
-	require.NoError(t, err)
-	err = os.Mkdir(tempDir, 0755)
-	require.NoError(t, err)
-
 	bucket, err := NewBucket(BucketOpts{
 		Location: PailLocal,
-		Path:     tempDir,
+		Path:     t.TempDir(),
```
johndaniels, line 16: Probably smart

johndaniels (approved):

PR #112, opened 2022-09-09
EVG-17670 precedence to build log lines

ybrill:
[EVG-17670](https://jira.mongodb.org/browse/EVG-17670)

For S3 backed logs, every time we pop a value from the heap the order of the iterators is swapped. Therefore, when the two timestamps are equal the min (which is j [here](https://github.com/evergreen-ci/logkeeper/blob/295b5bd446c5c993ac99a2b2fb12bfeb67d0d343/storage/log_iterator.go#L574)) will be the last popped iterator.

If we wanted to make it consistent with the db backed logs (which always [give precedence to build logs](https://github.com/evergreen-ci/logkeeper/blob/295b5bd446c5c993ac99a2b2fb12bfeb67d0d343/model/log.go#L358-L367)) we could stop using the heap (since we always only have two iterators) or we could do what this PR proposes.


ybrill:
I talked this over with @julianedwards and we decided not to make this change for now.

PR #114, opened 2022-09-14
EVG-17688: Remove DB writes

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-17688

This removes the DB writes, I refactored some of the code and cleaned up / added tests. Unfortunately, since we still have to fallback to the DB when reading older logs, I could not clean up everything.

I also checked in staging that we can write new logs, read logs living in S3, and read logs that only live in the DB.


views.go
```diff
@@ -698,75 +538,54 @@ func (lk *logKeeper) viewBucketLogs(r *http.Request, buildID string, testID stri
...
-		return nil, &apiError{Err: "finding build", code: http.StatusInternalServerError}
+		return nil, &apiError{Err: "finding build", code: http.StatusInternalServerError}, false
 	}
-
-	wg.Wait()
-	if testID != "" && pail.IsKeyNotFoundError(testErr) {
-		backgroundCancel()
-		return nil, &apiError{Err: "test not found", code: http.StatusNotFound}
+	if build == nil {
+		return nil, &apiError{Err: "build not found", code: http.StatusNotFound}, true
```
ybrill, line 570: was it not ideal to cancel the context when the build isn't found?

views.go
```diff
@@ -698,75 +538,54 @@ func (lk *logKeeper) viewBucketLogs(r *http.Request, buildID string, testID stri
...
 	return &logFetchResponse{
-		logLines:      logLines,
-		build:         build,
-		test:          test,
-		contextCancel: backgroundCancel,
-	}, nil
+		logLines: logLines,
+		build:    build,
+		test:     test,
+	}, nil, false
```
ybrill, line 588: should the contextCancel be removed from the `logFetchResponse` struct too?

views.go
```diff
@@ -496,7 +347,7 @@ func (lk *logKeeper) viewBuild(w http.ResponseWriter, r *http.Request) {
 	}{build, tests}, "base", "build.html")
 }
 
-func (lk *logKeeper) viewBucketBuild(r *http.Request, buildID string, shouldFallBack bool) (*model.Build, []model.Test, *apiError) {
+func (lk *logKeeper) viewBucketBuild(r *http.Request, buildID string) (*model.Build, []model.Test, *apiError) {
```
ybrill, line 350: could it make sense for this to also return a `fallback` boolean?

julianedwards: This doesn't really need it since we always want to fallback with a 404. When fetching logs only want to fall back if we get a 404 fetching the build, if we get a 404 for the test then we don't want to fall back since that we means we found the build in S3 but the test does not exist.

views.go
```diff
@@ -428,39 +303,15 @@ func (lk *logKeeper) appendLog(w http.ResponseWriter, r *http.Request) {
...
-		lk.logErrorf(r, "inserting DB log lines to test '%s' for build '%s': %v", buildID, testID, err)
-		lk.render.WriteJSON(w, http.StatusInternalServerError, apiError{Err: err.Error()})
-		return
-	}
-
-	wg.Wait()
-	if bucketErr != nil {
-		lk.logErrorf(r, "appending log lines to test '%s' for build '%s': %v", buildID, testID, bucketErr)
+	if err = lk.opts.Bucket.InsertLogChunks(r.Context(), buildID, testID, chunks); err != nil {
+		lk.logErrorf(r, "appending log lines to test '%s' for build '%s': %v", buildID, testID, err)
```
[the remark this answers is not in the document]
julianedwards: Good catch!

views.go
julianedwards: Yes, good catch!

views.go
julianedwards: I don't think we get much benefit from it since we are running these in parallel and if the build isn't found the other goroutines should return just as quick since there shouldn't be any test metadata or log lines for a build that does not exist.

views.go
julianedwards: Done.

ybrill (approved):

PR #116, opened 2022-09-20
EVG-17937 Add necessary headers for CORS requests

khelif96:
These headers are required to allow CORS requests from parsley. In the future we should validate the origins i opened EVG-17938 for that.

### DIFF
--- a/views.go
+++ b/views.go
@@ -30,9 +30,6 @@ const (
 
 func getCorsOrigins() []string {
 	origins := os.Getenv(corsOriginsEnvVariable)
-	if origins == "" {
-		return []string{}
-	}
 	return strings.Split(origins, ",")
 }
 
--- a/views.go
+++ b/views.go
@@ -312,6 +312,7 @@ func (lk *logKeeper) appendLog(w http.ResponseWriter, r *http.Request) {
 // GET /build/{build_id}
 
 func (lk *logKeeper) viewBuild(w http.ResponseWriter, r *http.Request) {
+	// TODO: EVG-17938 - Check this value against an allow list
 	requester := r.Header.Get("Origin")
 	w.Header().Add("Access-Control-Allow-Origin", requester)
 	w.Header().Add("Access-Control-Allow-Credentials", "true")
--- a/views.go
+++ b/views.go
@@ -30,6 +30,9 @@ const (
 
 func getCorsOrigins() []string {
 	origins := os.Getenv(corsOriginsEnvVariable)
+	if origins == "" {
+		return []string{}
+	}
 	return strings.Split(origins, ",")
 }
 
--- a/views.go
+++ b/views.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"html/template"
 	"net/http"
+	"os"
+	"strings"
 	"sync"
 	"time"
 
@@ -11,6 +13,7 @@ import (
 	"github.com/evergreen-ci/logkeeper/model"
 	"github.com/evergreen-ci/logkeeper/storage"
 	"github.com/evergreen-ci/render"
+	"github.com/evergreen-ci/utility"
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"github.com/mongodb/amboy"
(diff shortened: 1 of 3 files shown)


views.go
```diff
@@ -312,7 +312,11 @@ func (lk *logKeeper) appendLog(w http.ResponseWriter, r *http.Request) {
 // GET /build/{build_id}
 
 func (lk *logKeeper) viewBuild(w http.ResponseWriter, r *http.Request) {
-	w.Header().Add("Access-Control-Allow-Origin", "*")
+	// TODO: EVG-17938 - Check this value against an allow list
```
rtimmons, line 315: Yikes. Can we do..anything better than cors'ing the world?

khelif96: We already were cors'ing the world only difference is now we are cors'ing the world slightly less incorrectly 😞 

We need to add these headers in order to allow requests from parsley (The spiritual successor to lobster) to go through.  This wasn't necessary before because logkeeper hosted its own version of lobster and we didn't have to deal with cross origin requests. But now that we are hosting parsley on its own subdomain we need to allow these headers to bypass the CORS restrictions.

@johndaniels @ybrill is there an admin panel or something i can configure this too i don't mind adding the allow list here i just don't know if we have an established pattern for this.

khelif96: Done just opened a PR in kernel-tools https://github.com/10gen/kernel-tools/pull/347 to add the env variables.

* Added the allow list to this PR. 
* Checks the allow list against the requesting domain.  
* Added a fallback to the preexisting behavior just in case...

I also bumped the utility package to get `StringMatchesAnyRegex`

johndaniels (changes requested):

johndaniels:
New setup looks great, thanks for just picking this up @khelif96 !!

views.go
```diff
@@ -21,6 +24,30 @@ import (
 
 const maxLogBytes = 4 * 1024 * 1024 // 4 MB
 
+const (
+	corsOriginsEnvVariable = "LK_CORS_ORIGINS"
+)
+
+func getCorsOrigins() []string {
+	origins := os.Getenv(corsOriginsEnvVariable)
+	if origins == "" {
```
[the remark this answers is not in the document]
ybrill: I was just wondering about this.
[It looks like](https://pkg.go.dev/strings#Split) it will return "a slice of length 1 whose only element is" the empty string. Maybe? ([playground](https://go.dev/play/p/1FxLUg83DwC))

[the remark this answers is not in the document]
khelif96: In that case maybe i should revert  my commit?

views.go
```diff
@@ -21,6 +24,27 @@ import (
 
 const maxLogBytes = 4 * 1024 * 1024 // 4 MB
 
+const (
+	corsOriginsEnvVariable = "LK_CORS_ORIGINS"
+)
+
+func getCorsOrigins() []string {
```
ybrill, line 31: maybe do we want to do this in an init() function and compile the regexes in it?

khelif96: Is there any benefit to doing that vs what we currently do? We use this method on evergreen

views.go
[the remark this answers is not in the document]
ybrill: I mean, I think the empty string won't match anything. 
But it does seem like it would make it harder to understand what's happening.

views.go
```diff
@@ -21,6 +24,30 @@ import (
...
+
+func addCorsHeaders(w http.ResponseWriter, r *http.Request) {
+	requester := r.Header.Get("Origin")
+	// check if requester is in cors origins list
+	if utility.StringMatchesAnyRegex(requester, getCorsOrigins()) {
+		w.Header().Add("Access-Control-Allow-Origin", requester)
+		w.Header().Add("Access-Control-Allow-Credentials", "true")
+	} else {
+		// Maintain backwards compatibility with the old CORS header
+		w.Header().Add("Access-Control-Allow-Origin", "*")
```
ybrill, line 49: why is the "Access-Control-Allow-Credentials" header not necessary for other hosts?

views.go
```diff
@@ -21,6 +24,32 @@ import (
 
 const maxLogBytes = 4 * 1024 * 1024 // 4 MB
 
+const (
+	corsOriginsEnvVariable = "LK_CORS_ORIGINS"
+)
+
+var corsOrigins []string
+
+func init() {
```
ybrill, line 33: I think this would still be compiling the regexes for each request since `utility.StringMatchesAnyRegex` calls `regexp.MatchString` which Compiles the string regex.

views.go
khelif96: This was the old behavior of the app i opted to not change it.

ybrill (approved):

johndaniels (approved):

PR #119, opened 2022-10-11
EVG-18043 Add param to expose task metadata

khelif96:
This adds a query param that will return an object with a task and build id for a given build or build/test combo.

### DIFF
--- a/views.go
+++ b/views.go
@@ -386,11 +386,6 @@ func (lk *logKeeper) viewBucketBuild(r *http.Request, buildID string) (*model.Bu
 	return build, tests, nil
 }
 
-type BuildMetadata struct {
-	BuildID string
-	TaskID  string
-}
-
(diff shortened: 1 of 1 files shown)


views.go
```diff
@@ -408,6 +413,15 @@ func (lk *logKeeper) viewAllLogs(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if r.FormValue("metadata") == "true" {
+		buildMetadata := BuildMetadata{
+			BuildID: resp.build.ID,
+			TaskID:  resp.build.TaskID,
+		}
+		lk.render.WriteJSON(w, http.StatusOK, buildMetadata)
```
johndaniels, line 421: Just write the resp.build to json

khelif96: Done!

johndaniels (approved):

PR #122, opened 2022-10-19
EVG-18048: Update logkeeper to accept "Execution" as build/test metadata

opsnlops:
This patch allows for the inclusion of a `task_execution` parameter to be sent to logkeeper as metadata. It is an int.

### DIFF
--- a/views.go
+++ b/views.go
@@ -144,9 +144,10 @@ func (lk *logkeeper) createBuild(w http.ResponseWriter, r *http.Request) {
 	}
 
 	payload := struct {
-		Builder  string `json:"builder"`
-		BuildNum int    `json:"buildnum"`
-		TaskID   string `json:"task_id"`
+		Builder       string `json:"builder"`
+		BuildNum      int    `json:"buildnum"`
+		TaskID        string `json:"task_id"`
+		TaskExecution int    `json:"task_execution"`
 	}{}
 	if err := readJSON(r.Body, lk.opts.MaxRequestSize, &payload); err != nil {
 		lk.logErrorf(r, "bad request to create build: %s", err.Err)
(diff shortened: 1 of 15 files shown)


views.go
```diff
@@ -144,9 +144,10 @@ func (lk *logkeeper) createBuild(w http.ResponseWriter, r *http.Request) {
 	}
 
 	payload := struct {
-		Builder  string `json:"builder"`
-		BuildNum int    `json:"buildnum"`
-		TaskID   string `json:"task_id"`
+		Builder       string `json:"builder"`
+		BuildNum      int    `json:"buildnum"`
+		TaskID        string `json:"task_id"`
+		TaskExecution int    `json:"task_execution"`
```
julianedwards, line 150: Unfortunately this has to be `execution` since my change in resmoke passes in just `execution` instead of `task_exceution`. Sorry about that, I wasn't thinking :(.

julianedwards (changes requested):
Looks great! Could you update the tests in `views_tests.go` to make sure we are capturing the execution from the rest requests? Thanks!

views.go
opsnlops: @julianedwards Done!

opsnlops:
> Looks great! Could you update the tests in `views_tests.go` to make sure we are capturing the execution from the rest requests? Thanks!

Good catch, looks like that got missed in my megamerge. I'm on it!

opsnlops:
> > Looks great! Could you update the tests in `views_tests.go` to make sure we are capturing the execution from the rest requests? Thanks!
> 
> Good catch, looks like that got missed in my megamerge. I'm on it!

Done.

views_test.go
```diff
@@ -184,15 +190,17 @@ func TestCreateTest(t *testing.T) {
...
 	type payload struct {
-		TestFilename string `json:"test_filename"`
-		Command      string `json:"command"`
-		Phase        string `json:"phase"`
-		TaskID       string `json:"task_id"`
+		TestFilename  string `json:"test_filename"`
+		Command       string `json:"command"`
+		Phase         string `json:"phase"`
+		TaskID        string `json:"task_id"`
+		TaskExecution int    `json:"execution"`
```
julianedwards, line 197: Could you add an execution to the `NewTest` sub-test and check that it gets set there?

julianedwards (changes requested):

views_test.go
```diff
@@ -304,6 +314,7 @@ func TestAppendGlobalLog(t *testing.T) {
 		name               string
 		lk                 *logkeeper
 		buildID            string
+		taskExecution      int
```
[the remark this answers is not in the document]
opsnlops: Got this one!

views_test.go
opsnlops: Check again, please!

PR #125, opened 2022-11-02
EVG-17023: Log build and test log size stats to Splunk on every read request

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-17023

[Example of log stats in Splunk](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Dlogkeeper%20staging%20log%20size%20stats&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1667421540&latest=1667421580&sid=1667421598.6478103)

### DIFF
--- a/views.go
+++ b/views.go
@@ -21,7 +21,7 @@ import (
 
 const (
 	corsOriginsEnvVariable = "LK_CORS_ORIGINS"
-	maxLogBytes            = 4 * 1024 * 1024 // 4 MB
+	maxLogBytes            = 4 * bytesPerMB // 4 MB
 )
 
 var corsOrigins []string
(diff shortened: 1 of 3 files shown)


views.go
```diff
@@ -580,6 +567,64 @@ func (lk *logkeeper) viewBucketLogs(r *http.Request, buildID string, testID stri
...
+	var (
+		numLines    int
+		totalSize   int
+		maxLineSize int
+		minLineSize = maxLogBytes + len("\n")
+	)
+
+	var hasLines bool
+	for line := range resp.logLines {
+		hasLines = true
```
ybrill, line 580: do we want to write `lk.render.WriteJSON(w, http.StatusOK, nil)` for the `!hasLines` case?
Or at least to send a 200 response code?

julianedwards: It automatically returns a 200 and empty response, I checked in staging

ybrill: Ah nice!
I think [I found it](https://github.com/golang/go/blob/ba76567bc2500204432ed8a5cb28848410e74447/src/net/http/server.go#L1624-L1626).

ybrill (approved):

views.go
julianedwards: Good find!

PR #126, opened 2022-11-16
EVG-18291: Create a view build metadata route

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-18291

- Add a `metadata` parameter to the existing `/build/{build_id}` route that returns the build and test metadata as JSON.
- Return the test metadata instead of the build metadata for the `build/{build_id}/test/{test_id}` route.
- Add tests for all of the metadata routes.

### DIFF
--- a/views.go
+++ b/views.go
@@ -365,6 +365,15 @@ func (lk *logkeeper) viewBuild(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if r.FormValue("metadata") == "true" {
+		payload := struct {
+			model.Build
+			Tests []model.Test `json:"tests"`
+		}{*build, tests}
+		lk.render.WriteJSON(w, http.StatusOK, payload)
+		return
+	}
+
 	lk.render.WriteHTML(w, http.StatusOK, struct {
 		Build *model.Build
 		Tests []model.Test
@@ -481,7 +490,7 @@ func (lk *logkeeper) viewTestLogs(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if r.FormValue("metadata") == "true" {
-		lk.render.WriteJSON(w, http.StatusOK, resp.build)
+		lk.render.WriteJSON(w, http.StatusOK, resp.test)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


views.go
```diff
@@ -365,6 +365,15 @@ func (lk *logkeeper) viewBuild(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if r.FormValue("metadata") == "true" {
```
ybrill, line 368: if we have this do we still need the `metadata` option on the `build/{build_id}/all` route?

views.go
```diff
@@ -365,6 +365,15 @@ func (lk *logkeeper) viewBuild(w http.ResponseWriter, r *http.Request) {
...
+		payload := struct {
+			model.Build
```
ybrill, line 370: if this were a pointer to a build would the resulting JSON look different?

views_test.go
```diff
@@ -652,9 +653,28 @@ func TestViewBuild(t *testing.T) {
...
+				require.NoError(t, err)
+				tests, err := model.FindTestsForBuild(context.TODO(), buildID)
+				require.NoError(t, err)
+
+				expectedOut, err := json.MarshalIndent(struct {
+					model.Build
+					Tests []model.Test `json:"tests"`
+				}{*build, tests}, "", "  ")
+				require.NoError(t, err)
+				assert.Equal(t, expectedOut, resp.Body.Bytes())
```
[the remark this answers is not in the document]
julianedwards: We could, but then we would have to cast the values to strings. I think this looks cleaner.

views.go
julianedwards: Nope! The encoder ends up using reflection to get the underlying struct when it is a pointer. Pointers are really only useful when you want to omitempty with null values.

views.go
julianedwards: I was thinking that. Honestly I think it makes more sense for the `all` route to have the list of build + test metadata, but to make it consistent with this route I just put it here (since this is the "view build" route). If I could create it from scratch I would have just two routes: `build/{build_id}` and `build/{build_id}/test/{test_id}` which would show either all logs or the test logs, respectively, and both would accept a metadata query param.

julianedwards: I do think @khelif96 is using the `metadata` in the `all` route so we probably can't remove it.

ybrill (approved):

PR #129, opened 2022-11-28
EVG-18373 Point logkeeper job logs log links to parsley

khelif96:
This points all logkeeper job-logs page log links to parsley. 
Use env variables instead of hardcoding so we can have multiple environments. 

Depends on https://github.com/10gen/kernel-tools/pull/361
Do not merge until we are ready to launch Parsley!


templates/build.html
```diff
@@ -3,18 +3,24 @@
   <body>
     <div>
         <h2>
-          {{.Build.Builder}} - {{.Build.BuildNum}}       
+          {{.Build.Builder}} - {{.Build.BuildNum}}   
+          {{$evergreenURL := .EvergreenURL}}    
           {{if .Build.TaskID}}
-            (<a href="https://evergreen.mongodb.com/task/{{.Build.TaskID}}"> {{.Build.TaskID}}</a>)
+            (<a href="{{$evergreenURL}}/task/{{.Build.TaskID}}?redirect_spruce_users=true"> {{.Build.TaskID}}</a>)
```
ybrill, line 9: I don't understand why it's better to capture `.EvergreenURL` in a variable rather than using it directly from the context here. 
I have the same question about `.ParsleyURL` on line 13. I can understand why we need to capture it before the range, but couldn't that happen later, like how we capture the build in line 19?

ybrill (approved):

PR #130, opened 2022-12-27
EVG-18629 unmarshal log lines directly

ybrill:
[EVG-18629](https://jira.mongodb.org/browse/EVG-18629)

It appears that unmarshaling by using the decoder directly (as in [this example](https://pkg.go.dev/encoding/json#example-Decoder.Decode-Stream)) uses about half the time and an order of magnitude less memory/fewer allocations.
From inspection of the CPU profiles, (with sufficient handwaving) it could be the reason is that when we were calling the regular json.Unmarshal that called into the custom unmarshaler we were having it do double the work: the custom marshaler called back out to the json library to parse a new byte array for the inner array. At the same time the original call to the json library also needed to parse the inner array to keep track of where we were up to.

The before and after is striking:
before
```
go test -run=XXX -bench=. -benchmem  github.com/evergreen-ci/logkeeper/model
goos: darwin
goarch: arm64
pkg: github.com/evergreen-ci/logkeeper/model
BenchmarkReadLogJSONShort-10                        4248            283256 ns/op          210539 B/op       1233 allocs/op
BenchmarkReadLogJSONFewLongLines-10                    6         168763653 ns/op        139885874 B/op      2043 allocs/op
BenchmarkReadLogJSONManyShortLines-10                  5         244760442 ns/op        212711104 B/op   1200077 allocs/op
```

after
```
go test -run=XXX -bench=. -benchmem  github.com/evergreen-ci/logkeeper/model                                               
goos: darwin
goarch: arm64
pkg: github.com/evergreen-ci/logkeeper/model
BenchmarkReadLogJSONShort-10                        9276            127900 ns/op           50680 B/op        715 allocs/op
BenchmarkReadLogJSONFewLongLines-10                   14          76477333 ns/op        21030074 B/op        722 allocs/op
BenchmarkReadLogJSONManyShortLines-10                  9         120494009 ns/op        61725290 B/op     700036 allocs/op
```

It's entirely possible we could get even better, but this is definitely a good first step.

### DIFF
--- a/model/log_test.go
+++ b/model/log_test.go
@@ -1,6 +1,7 @@
 package model
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/evergreen-ci/logkeeper/env"
 	"github.com/evergreen-ci/logkeeper/testutil"
+	"github.com/evergreen-ci/utility"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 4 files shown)


ybrill:
When I looked at the heap profiles it occurred to me that the difference in memory is almost certainly because the original version reads the entire input into memory at once ([here](https://github.com/golang/go/blob/9123221ccf3c80c741ead5b6f2e960573b1676b9/src/encoding/json/stream.go#L63)) while the new version is only reading one line at a time.

model/log_test.go
```diff
@@ -422,3 +465,37 @@ func verifyDataStorage(t *testing.T, prefix string, expectedStorage expectedChun
...
 	assert.Equal(t, expectedStorage.body, string(actualChunkBody))
 }
+
+func benchmarkReadLogJSON(lineCount, lineSize int, b *testing.B) {
+	sampleJSON := makeJSONSample(lineCount, lineSize)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := UnmarshalLogJSON(bytes.NewReader(sampleJSON))
+		if err != nil {
+			b.Fatalf("unmarshal encountered error %s", err)
```
julianedwards, line 475: Nit: `'%s'` or `unmarshal encountered error: %s`

julianedwards (approved):
Great find! LGTM % nit

model/log_test.go
ybrill: done (both 😄 )

ybrill:
I wrote a naive parser that uses about as much memory as we do now but is many times as fast as the version in this PR.
```
go test -run=XXX -bench=. -benchmem  github.com/evergreen-ci/logkeeper/model
goos: darwin
goarch: arm64
pkg: github.com/evergreen-ci/logkeeper/model
BenchmarkReadLogJSONShort-10                               33945             34381 ns/op          118912 B/op        221 allocs/op
BenchmarkReadLogJSONFewLongLines-10                           82          13005919 ns/op        123129742 B/op       249 allocs/op
BenchmarkReadLogJSONManyShortLines-10                         43          27569861 ns/op        176579922 B/op    200071 allocs/op
BenchmarkReadLogJSONMaxLogSizeAverageLines-10                 25          46816113 ns/op        468875619 B/op     65608 allocs/op
BenchmarkReadLogJSONMaxLogSizeMaxLineSize-10                  27          40106256 ns/op        460578030 B/op        68 allocs/op
```

I can work on making it less naive if we don't mind parsing JSON ourselves in principle.

ybrill:
My plan is to hold off on the naive parser and see if we need it once this PR is deployed.

PR #131, opened 2023-01-11
EVG-1696 add execution to task link

ybrill:
[EVG-1696](https://jira.mongodb.org/browse/EVG-1696)

Add the task execution to the links on the build and test pages.

I was a bit unsure of what to do for builds that preceded https://github.com/evergreen-ci/logkeeper/commit/a5e8aebb5e4b4605ee11eddf7100855e6a784082 so they don't have an execution number. On the one hand, that was a few months back so it will be rare for a user to get bitten by it. On the other hand, it wasn't too hard to tell this is the case by making the build's task execution a pointer. Though I'll want to come back and clean it up after the 180 day TTL.

There's also a question of the builds created after we started to record executions but with an older version of resmoke that didn't send executions, so the recorded execution is 0. I think this is overthinking things.

### DIFF
--- a/model/build.go
+++ b/model/build.go
@@ -21,7 +21,7 @@ type Build struct {
 	Builder       string `json:"builder"`
 	BuildNum      int    `json:"buildnum"`
 	TaskID        string `json:"task_id"`
-	TaskExecution *int   `json:"execution"`
+	TaskExecution int    `json:"execution"`
 }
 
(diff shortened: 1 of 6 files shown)


ybrill:
We discussed that it wasn't worth handling builds from before the changes to record a build's task execution

opsnlops (approved):
LGTM! 👍🏻

opsnlops (approved):

PR #133, opened 2024-01-04
DEVPROD-3471: update deps

jayakasadev:
* upgrade to go 1.21
* update deps
* remove deprecated code


julianedwards:
evergreen retry

pprof.go
```diff
@@ -88,7 +85,7 @@ func profile(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "Could not enable CPU profiling: %s\n", err)
 		return
 	}
-	sleep(w, time.Duration(sec)*time.Second)
+	sleep(r, time.Duration(sec)*time.Second)
```
ybrill, line 88: I wonder: could these use the [time.Sleep](https://pkg.go.dev/time#Sleep) function?

jayakasadev:
```
func sleep(r *http.Request, d time.Duration) {
	ctx := r.Context()
	select {
	case <-time.After(d):
	case <-ctx.Done():
	}
}
```
sleep tries to terminate early if the request context closes

ybrill (approved):

pprof.go
ybrill: ah, that makes sense.
Thanks for explaining!

PR #135, opened 2024-01-18
DEVPROD-2551: add traces

jayakasadev:
* update go deps
* add otel trace spans
* add otel events
* update tests


logger.go
```diff
@@ -251,9 +251,9 @@ func sliceStats(sample, histogramBins []float64) message.Fields {
 	}
 	sort.Float64s(sample)
 
-	min := sample[0]
-	max := sample[len(sample)-1]
-	if len(histogramBins) == 0 || histogramBins[0] > min || histogramBins[len(histogramBins)-1] <= max {
+	minimum := sample[0]
+	maximum := sample[len(sample)-1]
```
jayakasadev, line 255: renamed to remove warnings

otel.go
```diff
@@ -20,30 +22,46 @@ const (
 	packageName = "github.com/evergreen-ci/logkeeper"
 )
 
-func (lk *logkeeper) initOtel(ctx context.Context) error {
-	if lk.opts.TraceCollectorEndpoint == "" {
+type Otel struct {
+	Tracer       *otelTrace.Tracer
+	OtelGrpcConn *grpc.ClientConn
+	Closers      []closerOp
+}
```
jayakasadev, line 29: made the method generic in order to instrument pprof service

pprof.go
```diff
@@ -22,23 +26,42 @@ import (
...
+	o, err := initOtel(ctx, "pprof", collectorEndpoint)
+	p.tracer = *o.Tracer
+	if err != nil {
+		grip.Error(message.WrapError(err, "error initializing otel"))
+	} else {
+		p.otelGrpcConn = o.OtelGrpcConn
+		p.closers = o.Closers
+	}
+	return &p
+}
```
jayakasadev, line 46: see above

main/logkeeper.go
```diff
@@ -45,15 +45,17 @@ func main() {
 	grip.EmergencyFatal(errors.Wrap(err, "getting bucket"))
 	grip.EmergencyFatal(errors.Wrap(env.SetBucket(&bucket), "setting bucket in env"))
 
+	tracer, err := logkeeper.InitOtel(ctx, *traceCollectorEndpoint)
+	defer logkeeper.Close(ctx)
```
jayakasadev, line 49: shared tracer

otel.go
```diff
@@ -20,30 +21,38 @@ const (
...
-		return errors.Wrap(err, "making host resource")
+		return tracer, errors.Wrap(err, "making host resource")
 	}
 
-	lk.otelGrpcConn, err = grpc.DialContext(ctx,
-		lk.opts.TraceCollectorEndpoint,
+	otelGrpcConn, err = grpc.DialContext(ctx,
+		collectorEndpoint,
 		grpc.WithTransportCredentials(credentials.NewTLS(nil)),
 	)
```
ybrill, line 47: if we don't need to hold on to the connection we could allow the otlptracegrpc.Client to manage its own connection. (like we do [here](https://github.com/evergreen-ci/evergreen/blob/352b79297bf6aa298879c37a98b93525d3311dcd/environment.go#L901))

views.go
```diff
@@ -139,29 +133,45 @@ func (lk *logkeeper) checkContentLength(r *http.Request) *apiError {
...
-		"error":   err,
-	})
+// To avoid expensive computations, check that span was sampled
+// before setting any attributes.
+func recordAttributes(ctx context.Context, attrs ...attribute.KeyValue) {
+	span := otelTrace.SpanFromContext(ctx)
+	if span.IsRecording() {
+		span.SetAttributes(attrs...)
+	}
 }
```
ybrill, line 143: are we doing sampling?

views.go
```diff
@@ -173,22 +183,30 @@ func (lk *logkeeper) createBuild(w http.ResponseWriter, r *http.Request) {
...
-		lk.logErrorf(r, "bad request to create build: %s", err.Err)
+		logErrorf(ctx, "bad request to create build: %s", err.Err)
 		lk.render.WriteJSON(w, err.code, err)
 		return
 	}
 
-	id, err := model.NewBuildID(payload.Builder, payload.BuildNum)
+	recordAttributes(
+		ctx,
+		attribute.String("builder", payload.Builder),
```
ybrill, line 193: It might be helpful to prefix all the attributes with something to make them easier to find.
In evergreen I had every attribute I set prefixed with `evergreen.<attribute>`.

model/build.go
```diff
@@ -58,7 +59,11 @@ func buildPrefix(buildID string) string {
 
 // NewBuildID generates a new build ID based on the hash of the given builder
 // and build number.
-func NewBuildID(builder string, buildNum int) (string, error) {
+func NewBuildID(ctx context.Context, builder string, buildNum int) (string, error) {
+	defer func() {
+		span := otelTrace.SpanFromContext(ctx)
+		span.AddEvent("NewBuildID")
+	}()
```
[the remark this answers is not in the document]
jayakasadev: no specific reason
I assumed that the method was lightweight enough that it didnt need a span

views.go
jayakasadev: should I use the same prefix for logkeeper?

otel.go
```diff
@@ -12,41 +13,50 @@ import (
...
+)
 
+func LoadTraceProvider(ctx context.Context, useInsecure bool, collectorEndpoint string, sampleRatio float64) {
+	if collectorEndpoint == "" {
+		return
+	}
 	r, err := serviceResource(ctx)
 	if err != nil {
-		return errors.Wrap(err, "making host resource")
+		grip.EmergencyFatal(errors.Wrap(err, "making host resource"))
```
ybrill, line 38: maybe it could just return an error? Is this definitely something we'd want to kill the application over?

otel.go
```diff
@@ -12,41 +13,50 @@ import (
...
 	}
-
-	lk.otelGrpcConn, err = grpc.DialContext(ctx,
-		lk.opts.TraceCollectorEndpoint,
-		grpc.WithTransportCredentials(credentials.NewTLS(nil)),
-	)
-	if err != nil {
-		return errors.Wrapf(err, "opening gRPC connection to '%s'", lk.opts.TraceCollectorEndpoint)
+	var opts []otlptracegrpc.Option
+	if useInsecure {
```
ybrill, line 41: When would we want to not use TLS?

otel.go
```diff
@@ -55,42 +65,49 @@ func (lk *logkeeper) initOtel(ctx context.Context) error {
...
 }
 
-func (lk *logkeeper) Close(ctx context.Context) {
-	catcher := grip.NewBasicCatcher()
-	for idx, closer := range lk.closers {
-		if closer.closerFn == nil {
-			continue
-		}
+func Close(ctx context.Context) {
+	closer.Do(func() {
```
ybrill, line 91: are we worried this will be called more than once?
And if it is called more than once would it be a problem?

otel.go
```diff
@@ -55,42 +65,49 @@ func (lk *logkeeper) initOtel(ctx context.Context) error {
...
 
-	return nil
+func initTracer(name string) otelTrace.Tracer {
+	var tracerName string
+	if useCustomName {
+		tracerName = fmt.Sprintf(packageName, name)
+	} else {
+		tracerName = defaultName
+	}
+	return otel.GetTracerProvider().Tracer(tracerName)
```
ybrill, line 87: The truth is, I think it works to have 
```go
var tracer = otel.GetTracerProvider().Tracer(<tracerName>)
```
At the top of a file. iirc the tracer will be initialized when the tracer provider is later initialized.

If this is true it could simplify things.

otel.go
```diff
@@ -12,41 +13,50 @@ import (
...
 
-	client := otlptracegrpc.NewClient(otlptracegrpc.WithGRPCConn(lk.otelGrpcConn))
 	traceExporter, err := otlptrace.New(ctx, client)
 	if err != nil {
-		return errors.Wrap(err, "initializing otel exporter")
+		grip.EmergencyFatal(errors.Wrap(err, "initializing otel exporter"))
 	}
 	tp := sdktrace.NewTracerProvider(
 		sdktrace.WithBatcher(traceExporter),
+		sdktrace.WithSampler(
```
ybrill, line 53: Do we need to do sampling in the app? Is the concern that we'll overwhelm the collector?
Separately, won't we always have the parent span since the entire trace originates on this one app server?

model/log.go
```diff
@@ -120,8 +124,10 @@ func (item *LogLineItem) OneSecondNewer(previousItem interface{}) bool {
 
 // DownloadLogLines returns log lines for a given build ID and test ID. If the
 // test ID is empty, this will return all logs lines in the build.
-func DownloadLogLines(ctx context.Context, buildID string, testID string) (chan *LogLineItem, error) {
-	buildKeys, err := getBuildKeys(ctx, buildID)
+func DownloadLogLines(ctx context.Context, tracer otelTrace.Tracer, buildID string, testID string) (chan *LogLineItem, error) {
+	_, span := tracer.Start(ctx, "DownloadLogLines")
```
ybrill, line 128: If we allow the returned ctx to override ctx it will allow the trace that continues with `getBuildKeys` to be threaded together.

model/build.go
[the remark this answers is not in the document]
ybrill: Ah, got it. fwiw span events and spans are billed the same, so it's totally a matter of taste.

views.go
ybrill: That works for me. Or `logkeeper.<whatever>`.

otel.go
jayakasadev: I figured it would be useful if we ever want to collect traces from devs' machines or other test conditions

otel.go
jayakasadev: I added it for safety, but since we explicitly call it, I removed it

jayakasadev: we may see errors if it is called twice. otel may complain that we are trying to close a closed connection

otel.go
jayakasadev: https://github.com/open-telemetry/opentelemetry-go/blob/main/internal/global/state.go#L42-L54

`GetTracerProvider` returns the default trace provider. In `LoadTraceProvider` we are explicitly setting the default trace provider, so we definitely want to init the tracer after `LoadTraceProvider`.

otel.go
jayakasadev: I set the sampler to make sampling decisions based on the parent span by default. If the remote/local parent span is being sampled, the sampler will also sample the span in `logkeeper`. If there isn't a parent span, the sampler will use the preset ratio for sampling.

otel.go
ybrill: But I think it maybe the way it works is that it returns a pointer to the trace provider (actually an interface) that then points to the initialized trace provider.

It worked for us in evergreen ([here](https://github.com/evergreen-ci/evergreen/blob/6d2d2cd8c71da036e032bd8714f397c791d5b89d/model/otel.go#L12) is an example).

otel.go
ybrill: But when will the parent ever not be sampled?

otel.go
```diff
@@ -12,41 +12,39 @@ import (
...
+		opts = append(opts, otlptracegrpc.WithInsecure())
 	}
+	opts = append(opts, otlptracegrpc.WithEndpoint(collectorEndpoint))
+	client := otlptracegrpc.NewClient(opts...)
 
-	client := otlptracegrpc.NewClient(otlptracegrpc.WithGRPCConn(lk.otelGrpcConn))
 	traceExporter, err := otlptrace.New(ctx, client)
 	if err != nil {
-		return errors.Wrap(err, "initializing otel exporter")
+		grip.EmergencyFatal(errors.Wrap(err, "initializing otel exporter"))
```
ybrill, line 37: flagging this `EmergencyFatal`

ybrill (approved):
Sorry for the delay

ybrill:
We should probably have refinery sampling in place before we deploy these changes.

ybrill (approved):

PR #136, opened 2024-02-06
DEVPROD-4532: set exec_timeout_secs to 1 hour

jayakasadev:
* set `exec_timeout_secs` to 1 hour

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -15,6 +15,8 @@ variables:
       - func: run-make
         vars: { target: "${task_name}" }
 
+exec_timeout_secs: 3600 # tasks will time out after 1 hour
+
 functions:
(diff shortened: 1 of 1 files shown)


.evg.yml
```diff
@@ -15,6 +15,8 @@ variables:
       - func: run-make
         vars: { target: "${task_name}" }
 
+exec_timeout_secs: 3600 # tasks will time out after 1 hour
```
julianedwards, line 18: Are these tasks ever going over an hour?

ybrill: It seems safe: looking at the waterfall all the versions I saw finished in ~12 minutes.

julianedwards (approved):

PR #137, opened 2024-02-15
DEVPROD-2551: add env vars for otel traces

jayakasadev:
* remove flags
* move flags to env vars except useInsecure because otel url is https
  * LK_COLLECTOR_ENDPOINT
  * LK_TRACE_RATIO

### DIFF
--- a/main/logkeeper.go
+++ b/main/logkeeper.go
@@ -46,8 +46,9 @@ func main() {
 		traceCollectorEndpoint = value
 	}
 	if value, ok := os.LookupEnv("LK_TRACE_RATIO"); ok {
-		if sampleRatio, err = strconv.ParseFloat(value, 64); err != nil {
-			grip.Warning(errors.Wrap(err, "using default trace ratio"))
+		var ratio float64
+		if ratio, err = strconv.ParseFloat(value, 64); err == nil {
+			sampleRatio = ratio
 		}
 	}
 
--- a/main/logkeeper.go
+++ b/main/logkeeper.go
@@ -38,17 +38,16 @@ func main() {
 
 	var (
 		err                    error
-		traceCollectorEndpoint string
-		sampleRatio            float64
+		traceCollectorEndpoint = ""
+		sampleRatio            = 1.0
 	)
 
 	if value, ok := os.LookupEnv("LK_COLLECTOR_ENDPOINT"); ok {
 		traceCollectorEndpoint = value
 	}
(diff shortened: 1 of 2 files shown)


otel.go
```diff
@@ -16,19 +16,15 @@ import (
...
 	r, err := serviceResource(ctx)
 	if err != nil {
 		return errors.Wrap(err, "making otel host resource")
 	}
-	var opts []otlptracegrpc.Option
-	if useInsecure {
-		opts = append(opts, otlptracegrpc.WithInsecure())
-	}
-	opts = append(opts, otlptracegrpc.WithEndpoint(collectorEndpoint))
+	opts := []otlptracegrpc.Option{otlptracegrpc.WithEndpoint(collectorEndpoint)}
```
ybrill, line 27: it looks like there's only one opt now. Could it make sense to just pass it to `otlptracegrpc.NewClient`?

main/logkeeper.go
```diff
@@ -29,15 +30,28 @@ func main() {
...
+	var (
+		err                    error
+		traceCollectorEndpoint string
+		sampleRatio            float64
+	)
+
+	if value, ok := os.LookupEnv("LK_COLLECTOR_ENDPOINT"); ok {
+		traceCollectorEndpoint = value
+	}
+	if value, ok := os.LookupEnv("LK_TRACE_RATIO"); ok {
```
ybrill, line 48: It seems like if `LK_TRACE_RATIO` isn't set in the environment the sample ratio will be 0. Is that what we want? I would have thought we'd want to default to 1.0.

main/logkeeper.go
```diff
@@ -29,15 +30,27 @@ func main() {
...
+		err                    error
+		traceCollectorEndpoint = ""
+		sampleRatio            = 1.0
+	)
+
+	if value, ok := os.LookupEnv("LK_COLLECTOR_ENDPOINT"); ok {
+		traceCollectorEndpoint = value
+	}
+	if value, ok := os.LookupEnv("LK_TRACE_RATIO"); ok {
+		if sampleRatio, err = strconv.ParseFloat(value, 64); err != nil {
```
ybrill, line 49: I'm not sure what `ParseFloat` will return in the case of error. It seems like it's probably going to set `sampleRatio` to 0 if there's an error.

jayakasadev: yeah the default is zero on error

ybrill (approved):

PR #140, opened 2024-06-21
DEVPROD-7162: Remove write API

julianedwards:
JIRA: https://jira.mongodb.org/browse/DEVPROD-7162

Stop allowing writes after completion of the [Consolidate Log Handling](https://jira.mongodb.org/browse/DEVPROD-75) project. Once the existing Logkeeper logs TTL, we will remove the app entirely.


ybrill (approved):
lgtm. 

There's a bunch of supporting code that could probably also go away, but it's probably not worth the effort of figuring out what it is given this whole repo is about to disappear.

It looks like there's still a bunch of writes happening, though. I checked one and it was still mongosync. You mentioned they still need to backport?

julianedwards:
>It looks like there's still a bunch of writes happening, though. I checked one and it was still mongosync. You mentioned they still need to backport?

No need for a backport, just local changes that need to rebase with main.

